PR #3, opened 2023-02-14
Uefi docker

PatrickRudolph:
Add docker files to handle UEFI builds.


MDr164 (approved):
lgtm, can fix the optional linting issues later.

PR #4, opened 2023-02-23
feat(Dockerfile): Update docker files

PatrickRudolph:
- Remove non root user, it breaks writing to mounted volumes
- Use WORKDIR ... RUN as suggested by the linter


MDr164 (approved):
More linter fixes are in #5 as well. That PR also reduces the time it takes to run the linter as it uses a smaller variant.

PR #6, opened 2023-02-23
Add Taskfile and Go binaries

MDr164:
Remaining TODOs: Automate invoking task via CI when doing semantic release. Evaluate either semanticversion or goreleaser actions for this repo. Tendency is to use the first.


MDr164:
I'll see if I can update this PR later today

MDr164:
I might revisit this at a later point

PR #10, opened 2023-05-05
(feat): Add edk2-stable202208 Dockerfile

PatrickRudolph:


MDr164:
Can you format the commit message according to the conventional commits guideline? I'll update the README with a link to that guideline as well. It's required to easily generate changelogs and define the scopes directly within commit messages. In this case it should be: `feat: add edk2-stable202208 Dockerfile`

MDr164 (approved):

PR #31, opened 2023-07-17
ci: run linter on every push and pull request

AtomicFS:


AtomicFS:
@MDr164 any chance of getting approving review?

MDr164 (approved):
Makes sense to always run the linters, though the megalinter needs to be slimmed down a bit so it runs faster.

PR #45, opened 2023-08-10
ci: make publishing conditional

AtomicFS:
Publish only on release or on main branch. This should keep the container registry clean.

### DIFF
--- a/.github/workflows/docker-build-and-test.yml
+++ b/.github/workflows/docker-build-and-test.yml
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   build:
-    name: Build and test
+    name: build_test_publish
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


AtomicFS:
Ready @MDr164

.github/workflows/docker-build-and-test.yml
```diff
@@ -36,8 +36,16 @@ jobs:
       - name: Install deps
         run: pip install -r ./.dagger-ci/daggerci/requirements.txt
       - uses: actions/checkout@v3
-      - name: Run Dagger pipeline
-        run: python .dagger-ci/daggerci/main.py -d ${{ matrix.dockerfile }} --publish
+      - name: Run dagger pipeline
+        run: |
+          if [[ "${GITHUB_EVENT_NAME}" == 'release' ]] || [[ "${GITHUB_REF}" == *'main' ]]; then
```
MDr164, line 41: I'm usually a fan of splitting jobs that are only supposed to run on a certain branch into a separate yaml and adapt the `on:` stanza but given that would cause a lot of duplication here it should be fine to have a condition here. And using the yaml builtin `if` (https://docs.github.com/en/actions/learn-github-actions/expressions#example-expression-in-an-if-conditional) is a bit too cursed as well.

MDr164 (approved):
lgtm

.github/workflows/docker-build-and-test.yml
AtomicFS: Yeah, I had the exact same dilema. I believe GitLab provide such option, that you can just reference another job, and change a thing or two very efficiently. But I have not found similar feature in GitHub :(

As for the yaml `if`, if does not have `else`, which would make things rather ugly, especially if we later decide to add more conditions (like schedule event for example).

This is not my proudest work, but should do the trick.

PR #49, opened 2023-08-14
tweak triggers for ci

AtomicFS:
- push to branch with existing pull request should no longer create duplicate workflows / actions for push and pull


AtomicFS:
Does not work and there is no easy way to make it work [How to trigger an action on push or pull request but not both?](https://github.com/orgs/community/discussions/26276)

PR #59, opened 2023-08-24
Feature/go and taskfile 2

AtomicFS:
More refactoring on the way


AtomicFS:
Also fixed pytest and added go test workflow

action.yml
```diff
@@ -10,49 +10,60 @@ inputs:
...
       For coreboot and linux this is a defconfig.
       For EDK2 this is a one-line file containing the build defines.
     required: true
-  platform:
+  output:
     description: |
-      *platform* specifies the DSC to use when building EDK2.
-      For example:
-        UefiPayloadPkg/UefiPayloadPkg.dsc
+      *output* specifies path (relative) to where place
```
MDr164, line 49: This sentence reads slightly odd. I'd change the word ordering to `*output* specifies the (relative) path where to place the produced files`

action/container/container.go
```diff
@@ -67,23 +68,42 @@ func Setup(ctx context.Context, client *dagger.Client, opts *SetupOpts) (*dagger
...
+			success, err = output.Export(
+				ctx,
+				artifact.HostPath,
+				dagger.FileExportOpts{AllowParentDirPath: true},
+			)
+		}
+
+		// Copy contents of containers artifacts directory to host
+		if err != nil || !success {
+			return fmt.Errorf("%w: %w: %s -> %s", errExportFailed, fmt.Errorf("%w", err), artifact.ContainerPath, artifact.HostPath)
```
MDr164, line 105:
```suggestion
			return fmt.Errorf("%w: %w: %s -> %s", errExportFailed, err, artifact.ContainerPath, artifact.HostPath)
```

action/recipes/coreboot.go
```diff
@@ -0,0 +1,69 @@
...
+			WithExec(buildSteps[step]).
+			Sync(ctx)
+		if err != nil {
+			return fmt.Errorf("coreboot build failed: %w", err)
+		}
+	}
+
+	// Extract artifacts
+	err = container.GetArtifacts(ctx, corebootContainer, artifacts)
+	if err != nil {
```
MDr164, line 64: You don't need to check the error here and return nil other wise as this is the end of the function. You can just return the result directly `return container.GetArtifacts(ctx, corebootContainer, artifacts)`

action/recipes/recipes.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+//===================
+// Universal options
+//===================
+
+var errRequiredOptionUndefined = errors.New("required option is undefined")
+
+type (
+	getValFunc func(string) string
```
MDr164, line 27: Currently the linter is rather lax, otherwise it'll complain here, that this line is not formatted right:
`type getValFunc func(string) string` as you'd normally only use braces when there are multiple type declarations back to back. (Just as in C less characters are preferred)

action/recipes/recipes.go
```diff
@@ -0,0 +1,159 @@
...
+			missing = append(missing, key)
+		}
+	}
+	if len(missing) > 0 {
+		return opts, fmt.Errorf("%w: %s", errRequiredOptionUndefined, strings.Join(missing, ", "))
+	}
+
+	// Check if sdk_version is URL, if not make it into URL defaulting to our containers
+	_, err := url.ParseRequestURI(opts.sdkVersion)
+	if err != nil {
```
MDr164, line 74: nit: Could be scoped a bit more narrow: `if _, err := ...; err != nil {`

MDr164 (commented):
Looks really good overall, especially the tests! I just got a few smaller nits but it's mostly good to go.

.github/workflows/go-test.yml
```diff
@@ -0,0 +1,31 @@
...
+  contents: read
+
+jobs:
+  go-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
AtomicFS: Copy-paste error, should have been `fetch-depth: 1` (which is the default)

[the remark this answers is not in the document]
AtomicFS: If I am not mistaken, we only need `fetch-depth: 0` in `lint.yml/commit-lint`, and everywhere else this can be removed. Maybe also `lint.yml/megalinter` but I am not sure about that

action.yml
AtomicFS: nevermind

.github/workflows/go-test.yml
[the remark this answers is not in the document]
MDr164: Currently only commit lint requires it as it needs access to the whole git history. Other candidates would be automatic release tools which need to see the whole git log in order to generate changelogs or tags. Megalinter shouldn't need it as well as it just operates on the code and not the git logs.

MDr164 (approved):

PR #61, opened 2023-09-01
Add linux build function

AtomicFS:
- moved content of `coreboot` building function into more generic `buildWithKernelBuildSystem` so that it can be re-used to build Linux kernel too
- added cross-compilation support
- added `architecture` into `action.yml` to allow cross-compilation
- tweaked `Taskfile.yml` to shuffle tests
- fixed problem with changing directories (test would change current working directory, but not change it back, which would break following test)
- add option to build container from Dockerfile for local testing


AtomicFS:
@MDr164 Please review approval after #62

action/container/container.go
```diff
@@ -46,8 +45,27 @@ func Setup(ctx context.Context, client *dagger.Client, opts *SetupOpts) (*dagger
 		return nil, errDirectoryInvalid
 	}
 
-	// Pull docker container
-	container := client.Container().From(opts.ContainerURL)
+	// Setup container either from URL or build from Dockerfile
+	var container *dagger.Container
+	if dockerfileDirectoryPath == "" {
+		// Use URL
+		fmt.Printf("Container setup: URL mode")
```
MDr164, line 52: If you want to use the `fmt` package for logging use either `fmt.Printf("MSG\n")` or if you don't need any format specifiers like in this line you can shortcut with `fmt.Println("MSG")`

action/recipes/linux_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/9elements/firmware-action/action/filesystem"
+	"github.com/Masterminds/semver"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLinux(t *testing.T) {
+	// This test is really slow (like 100 seconds)
+	/*
+		if testing.Short() {
+			t.Skip("skipping test in short mode")
```
MDr164, line 26: Any particular reason to comment out the skip in short mode?

action/recipes/coreboot_test.go
```diff
@@ -66,28 +68,31 @@ func TestCoreboot(t *testing.T) {
 		filepath.Join(pwd, fmt.Sprintf("../../tests/coreboot_%s/seabios.defconfig", corebootVersion)),
 		defconfigPath,
 	)
-	// ^^^ this relative path might be funky
+	//   ^^^ this relative path might be funky
```
MDr164, line 71: One thing we could do to make it more obvious to run the tests from the correct path is to add a message to the particular asserts that check errors related to missing files. So for example something like this:
```go
assert.NoErrorf(t, err, "encountered issue with missing files, is %s the root of the repo?", currentPath)
```

action/recipes/linux_test.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			common, err := commonGetOpts(getFunc)
+			assert.NoError(t, err)
+			linuxOpts := linuxOpts{}
+
+			// Change current working directory
+			//   create __tmp_files__ directory to store source-code of Linux Kernel
+			//   mostly useful for repeated local-run tests to save bandwidth and time
+			tmpFiles := filepath.Join(pwd, "__tmp_files__")
+			err = exec.Command("mkdir", "-p", tmpFiles).Run()
```
MDr164, line 100: Why not `os.MkdirAll()`?

action/recipes/linux_test.go
```diff
@@ -0,0 +1,169 @@
...
+			assert.NoError(t, err)
+			err = os.Chdir(tmpFiles)
+			assert.NoError(t, err)
+
+			// Download linux source code to __tmp_files__
+			var commands [][]string
+			if errors.Is(filesystem.CheckFileExists(fmt.Sprintf("linux-%s", linuxVersion.String())), os.ErrNotExist) {
+				commands = [][]string{
```
MDr164, line 108: I think for now we can leave these tests to just run a bunch of external commands, ideally the tests should not be dependent on the outside OS as we shouldn't assume the users environment if they want to reproduce the test results. Both `wget` and `unxz` are easy to do in Go natively, for the `gpg2` stuff we'd have to check if a library already exists (probably it does). But let's keep that in mind for a later release.

MDr164 (commented):
Quite a large amount of changes, will have to give it a more thorough round of review when I'm back. The comments I made occur on multiple places around the code, some are quick to change for other we might need to see if they are worth changing at all for now.

action/recipes/linux_test.go
AtomicFS: Oversight, my bad.

action/recipes/linux_test.go
AtomicFS: I forgot about it :D

action/recipes/coreboot_test.go
AtomicFS: Technically speaking, this should not pose a problem to the end user, only to us. And only in case we decide to move stuff around and the relative path will no longer match the directory tree.
But I agree that having helpful error message is better than comment.

Taskfile.yml
```diff
@@ -41,7 +41,7 @@ tasks:
     desc: Run tests
     dir: action
     cmds:
-      - go test {{.CLI_ARGS}} -covermode=count -coverprofile coverage.out ./...
+      - go test {{.CLI_ARGS}} -timeout 0 -shuffle=on -covermode=count -coverprofile coverage.out ./...
```
MDr164, line 44: Do we implement a timeout in the CI at the moment? By completely disabling the timeout you could potentially create a CI job that'll run for 24h without yielding a result in case of a faulty test. I'd rather prefer setting the timeout really high instead of disabling it completely, e.g. one hour or so

AtomicFS: Go has default timeout of 10 minutes ... found out the hard way.

AtomicFS: Also, it is kinda hard to gauge what timeout to set since on my PC it takes around 5 minutes to run these tests, but in CI it takes almost 20 minutes.

MDr164: I know, CI might take longer and 10min are already really short given for the complexity and time those tests take. That's why I proposed setting the timeout really high to be on the safe side but also not risk a never ending test to hog the CI. Maybe let's do 60min for now?

AtomicFS: OK, can do.

MDr164 (approved):
Todos will be tackled later as discussed, so lgtm

PR #62, opened 2023-09-01
add cross-compile toolchains to docker containers

AtomicFS:


AtomicFS:
I think this is review ready @MDr164

AtomicFS:
I changed the code to allow build container from Dockerfile during test to actually test my changes before merge (see #61 ). As it turns out, there were problems with the cross-compilation, but it should be now fixed.

MDr164 (approved):
lgtm

PR #66, opened 2023-09-08
Add edk2 build function

AtomicFS:


action/recipes/edk2.go
```diff
@@ -0,0 +1,68 @@
...
+		MountHostDir:      common.repoPath,
+		WorkdirContainer:  common.containerWorkDir,
+	}
+	myContainer, err := container.Setup(ctx, client, &containerOpts, dockerfileDirectoryPath)
+	if err != nil {
+		return err
+	}
+
+	// Copy over the defconfig file
+	// TODO ??? Is there even config to copy over?
```
MDr164, line 36: I think you already answered yourself in the action.yml file stating that this file only contains the build args, no?

AtomicFS: Yeah, I think so.

action/recipes/edk2_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+
+	pwd, err := os.Getwd()
+	assert.NoError(t, err)
+	defer os.Chdir(pwd) // nolint:errcheck
+
+	// Use "" if you want to test containers from github package registry
+	// Use "../../docker/linux" if you want to test containers built fresh from Dockerfile
```
MDr164, line 30: This comment needs to updated to say `../../docker/linux` I guess

action/recipes/recipes.go
```diff
@@ -74,7 +74,10 @@ func commonGetOpts(get getValFunc) (commonOpts, error) {
 	}
 
 	// Check if sdk_version is URL, if not make it into URL defaulting to our containers
-	if _, err := url.ParseRequestURI(opts.sdkVersion); err != nil {
+	//   WARNING: for url.ParseRequestURI is "edk2-stable202105:main" a valid URL for whatever reason ...
```
MDr164, line 77: The URI parsing follows the RFC 3986 and technically the mentioned string falls into that scheme even though it makes no sense in the context of using this as an FQDN. So while the string is a valid URI it is not a valid FQDN which you'll try to achieve. Though I think this approach here is good enough to also match for valid FQDNs.

MDr164 (dismissed):
Just some minor nits about inline comments, otherwise looks good

action/recipes/edk2_test.go
AtomicFS: ups :D

action/recipes/recipes.go
AtomicFS: I see

MDr164 (approved):
lgtm

PR #70, opened 2023-09-19
fix(action): bump version of NodeJS to v20

AtomicFS:


MDr164 (approved):
CI is happy so that means node20 works fine :+1:

PR #80, opened 2023-10-06
test(dagger): use taskfile, fix pytest

AtomicFS:


.github/workflows/pytest.yml
```diff
@@ -16,6 +16,9 @@ jobs:
   pytest:
     runs-on: ubuntu-latest
     steps:
+      - name: Install Taskfile
+        run: |
+          sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
```
MDr164, line 21: I don't really got anything against using their CI script but be aware that this makes it always pull the latest version of Task, so in case they update things in a breaking manner our CI might also fail. This is highly unlikely but there is an alternative: https://github.com/arduino/setup-task 
For now it makes no difference at all so I'm fine with it.

MDr164 (approved):

PR #81, opened 2023-10-19
fix(action): use relative paths for blobs in container

AtomicFS:
- for blobs the build scrips or makefile tries to make them into absolute path by pre-pending current working directory (so far observed only for FSP, but that might be because FSP was the first blob to be included)
- simple solution is to use just relative paths inside the container

... it is always the path ...


AtomicFS:
@MDr164 review pls?

MDr164 (approved):

PR #85, opened 2023-10-31
feature: automatic gcc detection

AtomicFS:
Add mechanism to automatically select GCC to use in compilation.

This is especially useful for EDK2.


MDr164 (approved):
I'm not 100% sure if having to inputs of the same type where one gets ignored is a big improvement for readability but I see why you opted for it so I'd say lgtm

AtomicFS:
Yeah, I was not sure if to have it for Linux too, since it is unused at the moment. But maybe we might need it in the future? I don't know.
I was also considering to add it to coreboot to keep things homogeneous, but coreboot has it's own tool-chain, so it is even more useless there.

PR #91, opened 2024-01-12
add documentation hosted in pages

### LINKED ISSUE #21: Document how to use this action
Document the usage and provide few examples.

AtomicFS:
fixes #21


MDr164:
As an idea: Can we link docs in our pkg.go.dev page as well? Most of the time this is where people look for initial function/method level documentation so linking to more context from there would make sense imo. For instance the changes to the README will already be reflected there so the general link will be available. My idea would be to extend that onto the text markers you created if that is possible. Otherwise good work adding more docs!

AtomicFS:
I am not sure if I understand what you mean. Posting a link to [https://pkg.go.dev](https://pkg.go.dev) is fine by me, although I have no clue how.

> My idea would be to extend that onto the text markers you created if that is possible.

Sorry but you completely lost me here. `extend that onto the text markers`?

> Otherwise good work adding more docs!

Thanks! I am still working on documenting how the docker containers with the fake docker compose work.

MDr164:
I suppose yes. I'm also not 100% sure right now which package doc gets rendered where but that looks about right. In addition the README will also be rendered so as you're putting the link there it will already link to the main docs page from the README here and the one at go.dev

AtomicFS:
Added the comment. I hope it is sufficient.

MDr164 (approved):

Issue #92, opened 2024-01-15
refactor edk2 builds


AtomicFS:
Current firmware-action for building `edk2` allows to build only payloads.

Refactor the code to allow builds of standalone `edk2` and `intel fsp` (re-opening issue #64).

PR #95, opened 2024-01-22
refactor edk to make it more universal

AtomicFS:
fixes issue #92


AtomicFS:
Initially I though that this approach would be impossible, but as it turns out it is.

In the example there is `source ./edksetup.sh; build ...`.
However in some cases it might be necessary to do additional things like `export` in the middle:
`source ./edksetup.sh; export SOMETHING='SOMETHING_ELSE'; build ...`

Initially I though this would not work, but turns out I was wrong.

MDr164 (approved):

PR #96, opened 2024-01-22
fix(docker): add cleanup commands

AtomicFS:
- turns out the .dockerignore files have no effect on the build
- add rm -rf commands to remove .git directories


MDr164 (commented):
Does it make such a big difference removing the git folders? Without them the build systems can't derive a git hash anymore if they need to and a shallow clone should not consume that much space after all.

AtomicFS:
True, shallow clone should be better in this case.

It does make a difference for the edk2 container, which has around 4GB, and those `.git` directories have almost 1GB combined. So there is potential for 25% saving.

AtomicFS:
Shallow clone for edk2 might be problematic. They don't use branches or tags like normal people to reference version, but rather use specific commits.

MDr164:
> Shallow clone for edk2 might be problematic. They don't use branches or tags like normal people to reference version, but rather use specific commits.

https://stackoverflow.com/a/43136160 apparently that can be achieved with a small workaround

AtomicFS:
Yeah, I just found that too ... am curious if it will work with github git server.

AtomicFS:
This should be better. I also tried `--shallow-submodules` but that gave me problems, so I removed it.

The `.git` directories still make up significant portion of the container, but this way we should save up few hundreds of MB.

MDr164 (approved):

PR #97, opened 2024-01-24
fix: guard queue with mutex in recipes.go

MDr164:
As the queue slice used by the DAG library inside a callback is created outside of the callback it is not concurrency safe. This issue became apparent as the library spawns goroutines for each callback which tripped the test runners.


AtomicFS (approved):
Perfect

PR #110, opened 2024-02-08
fix(docker): missing ifdtool and cbfstool in container

AtomicFS:
`ifdtool` and `cbfstool` are compiled in `toolchain` stage, but are not copied over to `final` stage.


AtomicFS:
@MDr164 review pls
Merge after #112

MDr164 (approved):

PR #134, opened 2024-02-14
ci: add cache to reduce stress on git hosting servers

AtomicFS:
I can see that sometimes the `git clone` fails on coreboot stuff. I assume this is because today this project has been just hammering down a lot of things thanks to the new dependencybot :D

Anyway, this should reduce the load. There is minor speed increase too, but minuscule.

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -36,9 +36,21 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: stable
+
+      - name: Restore cached coreboot repo
+        uses: actions/cache/restore@v3
+        with:
+          path: ./my_super_dooper_awesome_coreboot
+          key: coreboot-${{ matrix.coreboot-version }}
       - name: Clone coreboot repo
         run: |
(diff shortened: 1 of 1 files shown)


MDr164:
Do you know how much data this feeds into the cache? Afaik the cache is bound to the repo and not the org so we likely to fill up someone elses storage but would be interesting to know how much cache space this takes.

AtomicFS:
The neat part is that it will not store the cache over and over. Since we check-out a specific version, it does not change over time. As such, the key is rather constant across all runs and branches. When it tries to save a cache (in any other than 1st run) if fails because the cache already exists.

As for the sizes (**per version!**)
- coreboot `~50 MB`
- linux `~130 MB`
- edk2 `~658 MB`

The [limit is 10GB per repository](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

So edk2 could theoretically cause us to run out of cache memory.

AtomicFS:
https://github.com/9elements/firmware-action/actions/caches

PR #137, opened 2024-02-16
refactor logging pt1

### LINKED ISSUE #107: increase verbosity of firmware-action
firmware-action should output more information about what it is doing, especially on error.

AtomicFS:
- part 1: this one adds some features to aid logging
- fixes #107 
- requires:
  - #157
  - #147


action/container/container.go
```diff
@@ -23,66 +26,126 @@ var (
...
 }
 
+// Validate the data in struct
+func (opts SetupOpts) Validate() error {
+	pc, _, _, _ := runtime.Caller(0)
+	funcName := logging.TrimNameFunction(pc)
+
+	// None of the directories can be empty string
+	var err error
+	err = nil
```
MDr164, line 42: Not too sure if I like this pattern (other than being able to squash this line and the one above) as you keep overwriting the error object instead of joining it with something like errors.Join. So while you still log the errors you are only returning the last one in the chain.

action/container/container.go
```diff
@@ -23,66 +26,126 @@ var (
...
 	//   have the container uploaded into package registry
 
-	// None of the directories can be empty string
-	for _, val := range []string{opts.MountContainerDir, opts.MountHostDir, opts.WorkdirContainer} {
-		if val == "" {
-			return nil, errDirectoryNotSpecified
-		}
-	}
+	pc, _, _, _ := runtime.Caller(0)
+	funcName := logging.TrimNameFunction(pc)
```
MDr164, line 91: This feels like it could rather be part of a slog handler as it feels boilerplate-y to have this on top of every function.

action/container/container.go
AtomicFS: Oh yeah :D
Ups :D

AtomicFS: It was supposed to use `errors.Join`, my bad. With `errors.Join`, would you like this pattern?

action/container/container.go
AtomicFS: I was kinda hoping to avoid making handler :/ but the thought of having this on top of every function is less desirable. Valid point

AtomicFS:
@MDr164 now we have a custom-made log handler which adds location to the message :)
(also fixed that pesky `errors.Join` thing)

action/logging/pretty_handler.go
```diff
@@ -0,0 +1,139 @@
...
+// Handle handles the Record
+func (h *PrettyHandler) Handle(ctx context.Context, r slog.Record) error {
+	// Add location where slog.<LEVEL> has been called
+	pc, _, _, _ := runtime.Caller(3)
+	//   0 = logging.(*PrettyHandler).Handle
+	//   1 = slog.(*Logger).log
+	//   2 = slog.Debug || slog.Info || slog.Warn || slog.Error
+	//   3 = place where we call slog.<LEVEL>
+	funcName := TrimNameFunction(pc)
+	r.Add("location", funcName)
```
AtomicFS, line 93: I am not sure if this hard-coded value `runtime.Caller(3)` is OK, but I think it would be difficult to resolve on the fly. I suppose one way would be to iterate over `runtime.Callers`  and find the proper level with regex?

Since it should not change, maybe can stay this way. Also not sure how much overhead dynamically figuring out would cost.

AtomicFS:
Not sure why `golangci-lint` fails, works on my machine 🫠

action/logging/pretty_handler.go
MDr164: There is also `debug.Stack()` but I rarely ever use those methods manually so I'd have too look into the docs for those to figure out a smarter way to get the trace.

MDr164:
> Not sure why `golangci-lint` fails, works on my machine 🫠

I suspect CI and your machine got a different Go toolchain and linter version which don't match well here in CI.

AtomicFS:
Yeah, I guess, I run `1.56.1` and CI on `1.56.2`.

action/logging/pretty_handler.go
```diff
@@ -0,0 +1,139 @@
...
+	baseHandler slog.Handler
+	opts        *slog.HandlerOptions
+	mutex       *sync.Mutex
+	buffer      *bytes.Buffer
+	indent      bool
+	output      io.Writer
+}
+
+// NewPrettyHandler returns a PrettyHandler
+func NewPrettyHandler(output io.Writer, opts *slog.HandlerOptions, indent bool) *PrettyHandler {
```
MDr164, line 38: I'm not too happy with the option passing here. Having an option struct and two single options makes me wonder why we don't have a HandlerOptions struct here which inherits the HandlerOptions struct from slog. An even better solution imo would be to use the functional options pattern here as it allows setting same defaults without the user needing to know all input options. Also makes sanitization obsolete inside this function (e.g. checking the contents of opts fields against nil.

AtomicFS: Functional options pattern seems like a good option. Will do that.

MDr164 (commented):
Looks pretty solid now in my opinion

AtomicFS:
Cool, I will start working on the rest of the codebase then.

action/logging/logging.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"log/slog"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// ThisShouldNotHappenMessage contains a string which should be used in error message where the error justifies
+// opening an issue in issue tracker
+var ThisShouldNotHappenMessage = "this should not happen, please open an issue in issue tracker at https://github.com/9elements/firmware-action/issues"
```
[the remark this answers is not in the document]
AtomicFS: Could be constant

action/logging/pretty_handler.go
```diff
@@ -0,0 +1,228 @@
...
+	// Deal with concurrency
+	h.mutex.Lock()
+	defer func() {
+		h.buffer.Reset()
+		h.mutex.Unlock()
+	}()
+
+	// Use inner handler (JSONHandler) to compute JSON object
+	//   h.baseHandler is initialized with h.buffer as output
+	err := h.baseHandler.Handle(ctx, r)
```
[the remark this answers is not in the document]
AtomicFS: Yeah I guess, I am just not too fond of in-lined if statements. In my opinion it just hinders readability a bit. But it here it should be fine.

MDr164 (approved):
Looks pretty solid, let's hope I didn't miss anything crucial during review :sweat_smile:

AtomicFS:
Worst case scenario, we find a problem in production :D

PR #143, opened 2024-02-21
ci(action): further reduce load on CI

AtomicFS:
- trigger on 'merge_group' and 'push to main' only if changes in docker files, so that pull requests do not trigger so many jobs
- add weekly build (Sunday midnight)


AtomicFS:
mistake

PR #147, opened 2024-02-22
make containers interactive

### LINKED ISSUE #109: make Docker interactive on fail
It would be great be be able to connect into Docker container on failed build to debug it.

AtomicFS:
- on build failure open ssh server in the container and let user connect into it to debug the problem
- to enable this feature user has to pass argument `--interactive`
- into the containers add few editors (nano, vim), tools and utilities people would expect
- ssh login is with password, which is randomly generated

Fixes #109 

Testing will be pain, but I think there needs to be some. So I will take a look at it.

### DIFF
--- a/action/container/ssh.go
+++ b/action/container/ssh.go
@@ -1,0 +1,187 @@
+// SPDX-License-Identifier: MIT
+
+// Package container for dealing with containers via dagger
+package container
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"math/rand"
+	"strings"
+	"sync"
+
+	"dagger.io/dagger"
+)
+
+// ErrParseAddress is raised when containers IP address could not be parsed
+var ErrParseAddress = errors.New("could not parse address string")
+
+// OptsOpenSSH stores options for SSH tunnel for OpenSSH function
+type OptsOpenSSH struct {
+	WaitFunc  func()      // Waiting function holding container with SSH running
+	Password  string      // Filled in by OpenSSH function
+	IPv4      string      // Filled in by OpenSSH function
+	Port      string      // Filled in by OpenSSH function
+	MutexData *sync.Mutex // Mutex for modifying data in this struct
+
+	// We could do with single channel here, but for clarity and less mental overhead there are 2
+	TunnelClose chan (bool) // Channel to signal that SSH tunnel is ready
+	TunnelReady chan (bool) // Channel to signal that SSH tunnel is not longer needed and can be closed
+}
+
+// Wait calls WaitFunc
+func (s OptsOpenSSH) Wait() {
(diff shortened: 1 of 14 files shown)


AtomicFS:
@MDr164 review ready btw

action/container/ssh.go
```diff
@@ -0,0 +1,158 @@
...
+	defer sshServiceTunnel.Stop(ctx) // nolint:errcheck
+
+	// Get and print instructions on how to connect
+	sshAddress, err := sshServiceTunnel.Endpoint(ctx)
+	if err != nil {
+		fmt.Println("problem getting address")
+		return err
+	}
+	sshAddressSplit := strings.Split(sshAddress, ":")
+	opts.Port = sshAddressSplit[1]
```
MDr164, line 130: Why are you setting the port here out side the lock and three lines further down inside the mutex lock?

MDr164 (commented):
Just some minor things about the opts generated from the created tunnel. If we can guarantee that `sshAddressSplit` is always at least 2 elements large I'm fine with the current approach. Otherwise we should sanity check it to not do some out of bounds access by accident and causing a crash.

action/container/ssh.go
AtomicFS: that is some copy-paste error since I set the port twice :D

AtomicFS: I will make this into function (method of the struct) and add some security to it.

AtomicFS:
@MDr164 what about this?

action/container/ssh.go
```diff
@@ -0,0 +1,187 @@
...
+// Wait calls WaitFunc
+func (s OptsOpenSSH) Wait() {
+	s.WaitFunc()
+}
+
+// Address function parses provided string and populates IPv4 and Port (port defaults to 22 if not found)
+func (s *OptsOpenSSH) Address(address string) error {
+	s.MutexData.Lock()
+	var err error
+	err = nil
```
MDr164, line 43: You can drop this line, Go always initializes with the zero/nil value of the underlying datatype. So the line above is sufficient here.

AtomicFS:
Just found issue with the wait function. It test it works find, but when using manually, it fails to exit.

```
47: start /usr/sbin/sshd -D
47: > in service 6asqg9rb2ii4g
Connect into the container with:
  ssh root@127.0.0.1 -p 34667 -o PreferredAuthentications=password
Password is:
  n4sTE5VlEZCJV3Wh
SSH up and running

^Ctask: Signal received: "interrupt"
task: Failed to run task "build-edk2": exit status 130
```

AtomicFS:
Fixed, did not realize that writing into channel was also blocking,

MDr164 (approved):

PR #153, opened 2024-02-28
Dependabot/update dagger

AtomicFS:
Just to be save, I have merged #151 and #152 into single PR and temporarily enabled Dagger builds to actually test the changes.

Dagger changed from 0.9.11 to 0.10.0, wanted to test the changes just to be on save side, because I have some issues on my local machine. Probably caused by obsolete [dagger package in Arch Linux](https://archlinux.org/packages/extra/x86_64/dagger/) (I flagged it as out-of-date).

### DIFF
--- a/.dagger-ci/daggerci/requirements.txt
+++ b/.dagger-ci/daggerci/requirements.txt
@@ -1,5 +1,5 @@
 anyio~=4.3
-dagger-io~=0.9
+dagger-io~=0.10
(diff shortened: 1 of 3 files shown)


MDr164:
Looks like they only adding things in 0.10 and didn't change any existing functions that we use. So should be fine to merge.

MDr164 (approved):

PR #162, opened 2024-03-04
fix(docker): edk2 repositories were missing files

AtomicFS:
- when I tried to build edk2 stuff, repositoried in /tools were empty
- the reason is that I forgot to add 'git checkout'


MDr164 (approved):
I didn't know enough about EDK2's build system to notice that this was actually needed, my bad on prematurely approving the previous PR then.

AtomicFS:
It is not really about EDK2 build system, more like how git works. The fails is really mine, I should have testing it before making PR, rather than trusting random person on the internet.

PR #182, opened 2024-03-21
download executable instead of JIT compilation

### LINKED ISSUE #30: npm: word-wrap package vulnerable to Regular Expression Denial of Service
The original word-wrap project seems to be abandoned, see [pull request 33](https://github.com/jonschlinkert/word-wrap/pull/33).

Consider using a maintained forked mentioned in [comment in pull request 33](https://github.com/jonschlinkert/word-wrap/pull/33#issuecomment-1521663026).

### LINKED ISSUE #108: speed up GitHub action
As mentioned in issue #83 , the action takes additional time because the go program is compile on every run.

This could be mitigated by converting the action into docker action. This however brings problems in other CI/CD pipelines with **docker inside docker** structure. Although solvable, a bit inconvininet.

Another problem would be a **local runs**. Right now use can just execute `firmware-action` to do things. If the action is distributed as docker container, it would be rather cumbersome to use.

After some discussion, I think that the best option to solve this would be to do what people at golangci do - on CI action run pull binary from releases on GitHub and skip compilation.

- [x] Convert action from compile-on-demand into using [pre-compiled releases](https://github.com/golangci/golangci-lint-action/blob/e2315b67db0fd76598007338449373a42c572b25/src/install.ts#L116)
  - [x] #181

AtomicFS:
- until now when firmware-action was used, the go code was compiled on the fly in the CI
- instead download a executable from release assets
- fixes #108 
- fixes #30 
- this also removes all JavaScript code that we had in the repository


AtomicFS:
@MDr164  the examples are failing and I have no idea how does the JavaScript work. I am probably just too stupid to figure out how to import. Any clue what am I doing wrong?

MDr164:
> @MDr164 the examples are failing and I have no idea how does the JavaScript work. I am probably just too stupid to figure out how to import. Any clue what am I doing wrong?

I guess you'll have to run npm install so it fetches the deps inside the runners environment

AtomicFS:
fuck JavaScript

AtomicFS:
The one downside of this PR is that it does not run on Windows or Mac. I do not know if they have some `wget` and `tar -xvzf` equivalents.

action.yml
```diff
@@ -20,5 +20,46 @@ inputs:
...
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        wget "https://github.com/AtomicFS/firmware-action/releases/download/${{ steps.version.outputs.version }}/firmware-action_${{ runner.os }}_${{ steps.arch.outputs.arch }}.${{ steps.ext.outputs.ext }}"
+        tar -xvzf "firmware-action_${{ runner.os }}_${{ steps.arch.outputs.arch }}.${{ steps.ext.outputs.ext }}"
+    - name: fetch in Windows
+      if: runner.os == 'Windows'
+      shell: bash
+      run: |
+        echo "use Linux"
```
MDr164, line 56: Does wget and tar work the same in macOS as you're only checking for Windows? Generally it would be good to keep Windows Runner compatibility and as only tar and something to fetch is needed we could probably substitute it here with something more Windows-like.

AtomicFS: I fixed it now so that the action should work on all systems currently available - meaning it can download and run the executable on `Linux`, `Windows` and `MacOS`.

Running the `firmware-action` executable fails on `Windows` because of JSON validation, and fails on `MaOS` because Docker is not properly installed. On windows it is just matter of creting a new windows-specific config with windows paths probably, so that the JSON validator is satisfied. As for MacOS, I have tried to install Docker into the GitHub virtual machine, but with no success.

As it is right now, I am not going to fix these issues since we do not have any real use-case for them. However with these new fixes the functionality of JavaScript-based action should be preserved.

MDr164:
Looks good now, just wondering why there is the whole changelog being marked as added in this branch. Is that correct?

AtomicFS:
> Looks good now, just wondering why there is the whole changelog being marked as added in this branch. Is that correct?

As I was replacing the `npm` version control with `cocogitto`, I noticed in the manpage that it can also generate changelog. And I thought it would be a good thing to have a changelong right in the repository (to make it independent of GitHub).

MDr164:
Ah, I overlooked the small `Update changelog` step in the prepare-release file

MDr164 (approved):
Quite a large PR again so I hope I didn't overlook anything else in here.

AtomicFS:
Not that large. The interesting (function-changing) bits are under 200 lines.

MDr164:
> Not that large. The interesting (function-changing) bits are under 200 lines.

Everything above 100+ changes is large to me :smiling_face_with_tear:

PR #216, opened 2024-05-27
fix: return not checked in container/ssh.go

AtomicFS:
We only use the scan function as blocker to for user input. Content of said input is irrelevant. Hence we should be able to ignore any error I believe.


AtomicFS:
@MDr164 review pls

MDr164 (approved):
Looks like the possible errors returned by Scanln are not relevant to us as it's mostly about odd input stream ends that are not a newline or EOF character so I think it's fine.

PR #227, opened 2024-06-13
build(deps): run update on npm packages

AtomicFS:
Manually update `npm` packages. We have a blocking issue with security patches for `tar` and `braces` packages, and for whatever reason `dependabot` did not update them yet.


AtomicFS:
Bypassing protections because of blocking `gqlparser` dependency issue.

PR #231, opened 2024-06-17
Run docker container builds on release

AtomicFS:
OK, this is a bit cumbersome, but it should fix the problem then a new release does not trigger a build of Docker containers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,11 +16,13 @@ on:
       - closed
 
 permissions:
-  contents: write  # To create a new release
+  contents: read
 
 jobs:
-  release:
+  release-golang:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # To create a new release
     if: startsWith(github.event.pull_request.title, 'Release:') && github.event.pull_request.merged == true
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -66,3 +68,39 @@ jobs:
...
+
+  release-docker:
+    # There is a depth limit on workflows triggered by workflows to avoid infinite loops.
+    # This release workflow is triggered by merging a Pull Request made by release-prepare workflow.
+    #
+    # In addition the Docker build cannot really happen because of the `paths` filter in `on.push` event config.
+    # I do not want to change that because then we would build Docker containers all the time, which is long and unnecessary.
+    #
+    # And unfortunately a workflow cannot easily trigger another (Docker container builds) because of other GitHub limits.
+    # However it is possible to use API to trigger a workflow. That is what this job is for.
```
MDr164, line 80: I guess we're close to finding every GitHub Actions edgecase with this repo. I'll give it a more detailed review later but so far looks good (for a necessary workaround).

AtomicFS: Yeah, I have a talent for finding edge-cases. Over-complicated GitHub design always gets in the way.

One of the problems is that `paths` in this example applies to both `branches` and `tags`:
```yaml
on:
  push:
    branches: ['main']
    paths: ['docker/**']
    tags: ['v*']
```

I really wanted to do something like:
```yaml
on:
  push:
    branches: ['main']
    paths: ['docker/**']
  push:
    tags: ['v*']
```
But that is not possible.

MDr164: Maybe splitting up the YAMLs with different triggers could get that done but I'm not sure if that looks any cleaner in the end compared to the short API call script here

AtomicFS: Then there would be a lot of copy-paste code.

MDr164 (approved):

PR #232, opened 2024-06-18
feat(action): add option to ignore missing blob files

### LINKED ISSUE #184: Allow `ifdtool_entries` to skip on missing file
Add option to skip item in `ifdtool_entries` if the file is missing.

AtomicFS:
fixes #184 

It can be handy sometimes to ignore if a blob is defined, but the actual file is missing. For example for a matrix builds where one variation needs the blob and another one does not.

### DIFF
--- a/action/recipes/stitching_test.go
+++ b/action/recipes/stitching_test.go
@@ -141,6 +141,12 @@ type makeFile struct {
 }
 
 func (base makeFile) MakeMe() error {
+	log.Printf("Path:       %s", base.Path)
+	log.Printf("Content:    %s", base.Content)
+	log.Printf("SourcePath: %s", base.SourcePath)
+	pwd, _ := os.Getwd()	// nolint:errcheck
+	log.Printf("PWD:        %s", pwd)
+
 	// If file does not exist, make it
 	if _, err := os.Stat(base.Path); os.IsNotExist(err) {
 		if base.Content != "" {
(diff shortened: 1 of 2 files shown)


action/recipes/stitching_test.go
```diff
@@ -141,6 +141,12 @@ type makeFile struct {
 }
 
 func (base makeFile) MakeMe() error {
+	log.Printf("Path:       %s", base.Path)
+	log.Printf("Content:    %s", base.Content)
+	log.Printf("SourcePath: %s", base.SourcePath)
+	pwd, _ := os.Getwd()
```
MDr164, line 147: I'm surprised the linter didn't complain here about the ignored error. It's totally fine here to drop it but still odd that you didn't need `nolint:errcheck`

action/recipes/stitching_test.go
```diff
@@ -141,6 +141,12 @@ type makeFile struct {
 }
 
 func (base makeFile) MakeMe() error {
+	log.Printf("Path:       %s", base.Path)
```
MDr164, line 144: I'm not fully in the loop in regards of logging. Did we switch to slog or is log still the way to go?

AtomicFS: Normally we use `slog`, but in tests we use `log`.

action/recipes/stitching_test.go
AtomicFS: True, not sure why that is, so I added `nolint:errcheck` to future-proof this.

action/recipes/stitching_test.go
MDr164: Totally overlooked that this is a test file. It makes no big difference but maybe we should use `t.Log()` but that would require this function to also take `*testing.T` as an input parameter and might be out of scope for this PR.

AtomicFS: `t.Log()` might be better than `slog`. I will put this into my todo list with low priority.

MDr164 (approved):

PR #264, opened 2024-07-15
fix(ci): remove on submitted review trigger

AtomicFS:
It is just causing doubling of CIs of each PR.


AtomicFS:
Never mind, it is still needed :(

PR #265, opened 2024-07-15
fix(ci): skip on all reviews except when the release PR

AtomicFS:
- it is just doubling number of CIs of each PR
- we have this only because the release PR does not trigger corectly workflows, because we have too many nested triggers I think (github has a limit to prevent infinite loops)
- this fix should shopefully skip PR review trigger when the PR is not made by the github-actions[bot] which we use for releases

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -29,6 +29,8 @@ jobs:
     strategy:
       matrix:
         coreboot-version: ['4.19', '4.20.1', '4.21', '24.02']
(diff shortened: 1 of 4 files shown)


MDr164 (approved):
I wonder if it makes more sense to a rule into the PR review trigger but given that you do a check on the actor I doubt it's possible so I guess this should be fine

AtomicFS:
> I wonder if it makes more sense to a rule into the PR review trigger but given that you do a check on the actor I doubt it's possible so I guess this should be fine

I was looking into that but the `review` trigger is extremely limited unfortunately. This was the simplest solution I could come up with.

PR #267, opened 2024-07-15
ci(docker): add first u-root container v0.14.0

### LINKED ISSUE #260: containers for u-root do not have sshd
To get `--interactive` working, we need `sshd` installed in the container. So we have to make our containers anyway.

AtomicFS:
fixes #260 

When I added the container for the first time I was seeing some validation errors. So just to make sure it is actual syntax issue in `compose.yaml` and not some python error I have also added extra validation step. It is unnecessary, but does not slow anything down, so I figured why not.

### DIFF
--- a/tests/test_uroot.sh
+++ b/tests/test_uroot.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+# Clone repo
+git clone --branch "${VERIFICATION_TEST_UROOT_VERSION}" --depth 1 https://github.com/u-root/u-root.git
+cd u-root
+
+# Make
+go build
+./u-root core boot
+
--- a/.github/workflows/docker-build-and-test.yml
+++ b/.github/workflows/docker-build-and-test.yml
@@ -81,7 +81,8 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
+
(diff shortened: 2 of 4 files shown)


tests/test_uroot.sh
```diff
@@ -0,0 +1,17 @@
...
+set -Eeuo pipefail
+
+# Clone repo
+git clone --branch "${VERIFICATION_TEST_UROOT_VERSION}" --depth 1 https://github.com/u-root/u-root.git
+cd u-root
+
+# Make
+go build
+./u-root \
+	-defaultsh gosh \
```
MDr164, line 12: Is this actually still needed? Not sure when they made gosh the default upstream.

tests/test_uroot.sh
```diff
@@ -0,0 +1,17 @@
...
+	-o initramfs.cpio \
+	boot coreboot-app \
+	./cmds/core/* \
+	./cmds/boot/*
```
MDr164, line 16: Shouldn't this already be part of the `boot` preset?

tests/test_uroot.sh
AtomicFS: I don't know, I just blindly copied stuff from CodingVoid.

tests/test_uroot.sh
AtomicFS: Yeah, when I look at `u-root`'s `README` I see that `u-root core boot` should be enough.

PR #270, opened 2024-07-16
feat(action): add InputFiles and InputDirs options

### LINKED ISSUE #261: Add options for input files and directories
For multi-stage builds, it would be great to be able to specify list of `input_files` and `input_directories`, just like right now we have `container_output_files` and `container_output_dirs`, only in reverse.

Related to #196

AtomicFS:
fixes #261


action/container/container.go
```diff
@@ -117,13 +131,47 @@ func Setup(ctx context.Context, client *dagger.Client, opts *SetupOpts, dockerfi
...
 			client.Host().Directory(opts.MountHostDir)).
-		WithWorkdir(opts.WorkdirContainer).
-		Sync(ctx)
+		WithWorkdir(opts.WorkdirContainer)
+
+	// Get current working directory
+	pwd, err := os.Getwd()
+	if err != nil {
+		slog.Error(
+			"Could not get working directory, should not happen",
```
MDr164, line 145: Isn't the "should not happen" duplicate with the suggestion message?

AtomicFS: Now that you mention it ... yes it is.

action/container/container_test.go
```diff
@@ -55,48 +55,102 @@ func TestSetup(t *testing.T) {
...
+				ContainerURL:      "ubuntu:latest",
+				MountContainerDir: "/src",
+				MountHostDir:      t.TempDir(),
+				WorkdirContainer:  "/src",
+				InputDirs:         []string{"test/"},
+			},
+			wantErr:       errDirectoryNotSpecified,
+			lsContains:    []string{},
+			lsNotContains: []string{},
 		},
```
AtomicFS, line 110: We also need to test the input directories.

action/recipes/config.go
```diff
@@ -69,6 +69,17 @@ type CommonOpts struct {
 	//     ├── coreboot.rom
 	//     └── defconfig
 	OutputDir string `json:"output_dir" validate:"required,dirpath"`
+
+	// Specifies the path to directory where to place input files and directories inside container.
+	//   Directories listed in ContainerInputDirs and files listed in ContainerInputFiles
+	//   will be copied there.
+	ContainerInputDir string `json:"container_input_dir" validate:"dirpath"`
```
MDr164, line 76: The naming of `InputDir` vs `InputDirs` feels a bit confusing to me, given that they don't mean the same thing and one isn't just the plural form. Maybe name the first one `ContainerInputTargetDir`? Otherwise I'm open to other suggestions but the current naming feels to confusing.

AtomicFS: You have a valid point, I will change this

AtomicFS: I have changed it so that it is more coherent to the existing `ContainerOutputDirs`, `ContainerOutputFiles` and `OutputDir` settings.

And just to make sure I have added more documentation.

action/recipes/config.go
```diff
@@ -63,12 +63,47 @@ type CommonOpts struct {
...
+	//   Directories listed in ContainerInputDirs and files listed in ContainerInputFiles
+	//   will be copied there.
+	// Example:
+	//   Following setting:
+	//     InputDirs = []string{"cofig-files/"}
+	//     InputFiles = []string{"README.md", "Taskfile.yml"}
+	//     ContainerInputDir = "myInput"
+	//   Will result in following structure being copied into the container:
+	//     myInput/
+	//     ├── cofig-files/
```
MDr164, line 89: `cofig` (linter also found that one)

action/container/container.go
```diff
@@ -25,10 +26,13 @@ var (
...
 type SetupOpts struct {
-	ContainerURL      string // URL or name of docker container
-	MountHostDir      string // Directory from host to mount into container
-	MountContainerDir string // Where to mount ^^^ host directory inside container
-	WorkdirContainer  string // Workdir of the container, specified by GITHUB_WORKSPACE environment variable
+	ContainerURL      string   // URL or name of docker container
+	MountHostDir      string   // Directory from host to mount into container
+	MountContainerDir string   // Where to mount ^^^ host directory inside container
+	WorkdirContainer  string   // Workdir of the container, specified by GITHUB_WORKSPACE environment variable
+	InputDir          string   // Directory for input files
```
MDr164, line 33: Overlooked this one here. Might be good to rename in the same notion to avoid confusion.

AtomicFS: Not sure what you mean. Is the current state unsatisfactory?

| Configuration option   | Direction            |
|:-----------------------|:--------------------:|
| RepoPath               | Host  --> Container  |
|                        |                      |
| ContainerOutputDirs    | Host <--  Container  |
| ContainerOutputFiles   | Host <--  Container  |
| OutputDir              | Host <--  Container  |
|                        |                      |
| InputDirs              | Host  --> Container  |
| InputFiles             | Host  --> Container  |
| ContainerInputDir      | Host  --> Container  |

AtomicFS: Sorry, I see what you mean now

AtomicFS:
Well, with removal of JIT compilation I now successfully broke the `example` jobs :D

Changes to the golang code do not appear in the jobs anymore because it download older pre-compiled binary.

.github/workflows/example.yml
```diff
@@ -27,9 +27,27 @@ permissions:
...
+  changes:
+    runs-on: ubuntu-latest
+    # Required permissions
+    permissions:
+      pull-requests: read
+    outputs:
+      action: ${{ steps.filter.outputs.action }}
+    steps:
+      # For pull requests it's not necessary to checkout the code
+      - uses: dorny/paths-filter@v3
```
MDr164, line 40: What's the benefit of this instead of using the built-in path triggers i GitHub?

AtomicFS: This is to decide whether or not to use pre-compiled binary executable from releases, or if a freshly-compiled binary is needed to actually tests the changes.

For example if we make changes in `action/` golang code, they would not be tested if the executable is downloaded from latest release. In the case we have to compile on the fly.

MDr164: Yes, I got that. I was just wondering what the benefit of doing this via a thirdparty action is vs the built-in `paths` filter: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request

AtomicFS: How could we use that? It is a trigger.

We could use it but then we would have to copy-paste the workflow - one for downloading from releases and one for compilation.

MDr164: Ah so you want to make certain steps optional without having to split the yaml and potentially duplicate code. Got it.

MDr164 (dismissed):

.github/workflows/example.yml
AtomicFS: Yes, exatly. I have added `input` to the action called `compile` which switches between compilation mode and download mode. This is where I feed in the boolean output of the `dorny/paths-filter`.

MDr164:
I do see a new minor issue looking at the checks: We currently require all the important tests to pass. Even if they are skipped due to not being necessary. We could likely just loosen the requirement for this repo in the settings, no code change necessary.

AtomicFS:
> I do see a new minor issue looking at the checks: We currently require all the important tests to pass. Even if they are skipped due to not being necessary. We could likely just loosen the requirement for this repo in the settings, no code change necessary.

Yeah, I just did loosen it up. But there is still a problem with the filter action ... for whatever reason it fails to see the repository.

```
Run dorny/paths-filter@v3
Get current git ref
  /usr/bin/git branch --show-current
  fatal: not a git repository (or any of the parent directories): .git
Error: The process '/usr/bin/git' failed with exit code 128
```

MDr164:
> > I do see a new minor issue looking at the checks: We currently require all the important tests to pass. Even if they are skipped due to not being necessary. We could likely just loosen the requirement for this repo in the settings, no code change necessary.
> 
> Yeah, I just did loosen it up. But there is still a problem with the filter action ... for whatever reason it fails to see the repository.
> 
> ```
> Run dorny/paths-filter@v3
> Get current git ref
>   /usr/bin/git branch --show-current
>   fatal: not a git repository (or any of the parent directories): .git
> Error: The process '/usr/bin/git' failed with exit code 128
> ```

Maybe the checkout you left out wasn't unnecessary after all? But even then it's weird that it worked previously.

AtomicFS:
> > > I do see a new minor issue looking at the checks: We currently require all the important tests to pass. Even if they are skipped due to not being necessary. We could likely just loosen the requirement for this repo in the settings, no code change necessary.
> > 
> > 
> > Yeah, I just did loosen it up. But there is still a problem with the filter action ... for whatever reason it fails to see the repository.
> > ```
> > Run dorny/paths-filter@v3
> > Get current git ref
> >   /usr/bin/git branch --show-current
> >   fatal: not a git repository (or any of the parent directories): .git
> > Error: The process '/usr/bin/git' failed with exit code 128
> > ```
> 
> Maybe the checkout you left out wasn't unnecessary after all? But even then it's weird that it worked previously.

Yeah, my exact thoughts. I think it is unnecessary only for PR's (`# For pull requests it's not necessary to checkout the code`) but it is still required for merge checks.

MDr164 (approved):

PR #277, opened 2024-07-25
Issue/166

### LINKED ISSUE #166: Resolve if output direcotry already exists
In the `JSON` user specifies `output_dir`. At the moment the content just gets over-written, which is not ideal.

It would be preferable to rename the old directory.

### LINKED ISSUE #196: Weird error when building non-recursivelly and previous stages are missing
You get some weird, undecipherable error when you try to build non-recursively a module while files from required modules are missing.

AtomicFS:
fixes #166 
fixes #196

### DIFF
--- a/action/main.go
+++ b/action/main.go
@@ -124,7 +124,7 @@ submodule_out:
 	}
 
 	// Lets build stuff
-	_, err = recipes.Build(
+	results, err := recipes.Build(
 		ctx,
 		CLI.Build.Target,
 		CLI.Build.Recursive,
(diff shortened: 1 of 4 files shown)


MDr164:
> Any suggestions?

I don't have any issues with having newline characters in the json output as those are escaped properly. So personally this looks fine visually.

AtomicFS:
If you think visually it is OK, then lets keep it as is for now.

action/main.go
```diff
@@ -124,16 +124,33 @@ submodule_out:
...
+		} else if errors.Is(item.BuildResult, recipes.ErrBuildSkipped) {
+			result = "Skipped"
+		} else {
+			result = "Fail"
+		}
+		summary = fmt.Sprintf("%s\n%s: %s", summary, item.Name, result)
+	}
+	slog.Info(summary)
+
 	if err == nil {
```
MDr164, line 151: This conditional here at the end is not really the 'Go way' to do it and I also don't know why I haven't noticed this before (probably because the check is so small): Normally we just test if `err != nil` and only branch out on error conditions and simply return nil at the functions end. Given that there is no extra logic other than this in the conditional block it makes no difference but for the future I'd also recommend having the 'happy path' in indentation level 0 and every error path inside a conditional.

AtomicFS: Yeah, than if statement is not the proudest piece of code, when looking at it now. The `return nil` is completely unnecessary.

AtomicFS: Oh, now I see, it is a bit of leftover from previous commits. [This was the original](https://github.com/9elements/firmware-action/commit/ee19f2945403e99e093600fe3d5411d93abffd5a).

AtomicFS: I just removed the `return nil`. I hope that now it is more acceptable.

MDr164 (approved):

PR #287, opened 2024-07-30
fix(release): remove leftover javascript stuff

AtomicFS:
Yet another release fix


MDr164 (dismissed):
Piping sed into sed is also interesting to see but I guess it's fine for that purpose

AtomicFS:
Ohhhh ... it is rather annoying how the release workflows are so often broken :/

PR #292, opened 2024-07-31
docs: add CODEOWNERS file

AtomicFS:
I think it would be handy, and super nice if / when more people will start working on this project.


MDr164:
I would even go as far and put the whole cybersec team as code owners instead of just us two. But on a second thought that will make them be tagged as reviewers on PRs. Let's just leave it like this then :+1:

AtomicFS:
I was also thinking about that but came to the same conclusion about tagging everyone in all PRs in the worst-case scenario, or tagging randomly selected lucky person in the best case scenario.

PR #293, opened 2024-07-31
fix(action): remove unnecessary apostrophes

AtomicFS:
We are getting some failures when the action is trying to download the executable from releases.

Right now it just downloads a small text file with `Not found` written inside it. This is because of the apostrophes surrounding the `firmware-actin` version in the URL.

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -51,7 +51,7 @@ jobs:
           sed -i -E "s/SEMVER: .*/SEMVER: '${{ steps.update_version.outputs.version }}'/g" Taskfile.yml
       - name: Update action.yml
         run: |
-          sed -i -E "s/version=v[a-zA-Z0-9\.]+/version='${{ steps.update_version.outputs.version }}'/g" action.yml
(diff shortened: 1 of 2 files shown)


MDr164:
Is this the only place where this was done?

AtomicFS:
> Is this the only place where this was done?

I hope so

AtomicFS:
No, you are right, found another place in `release-prepare.yml` workflow.

.github/workflows/release-prepare.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           sed -i -E "s/SEMVER: .*/SEMVER: '${{ steps.update_version.outputs.version }}'/g" Taskfile.yml
```
MDr164, line 51: My sed-fu isn't the greatest but would those inner quotes also be optional here?

AtomicFS: These are there because of taskfile syntax. Desired output is this:
```yaml
SEMVER: 'v0.4.0'
```

MDr164 (approved):

PR #295, opened 2024-07-31
fix(ci): jobs are canceled on submitted review

AtomicFS:
- we are using pull_request_review.submitted trigger to have CIs running for release PRs and releases
- the issue is that when someone submits a review on ordinary PR, it creates a bunch of jobs which are immediately skipped, but they still cancel already running jobs because they are in the same group

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -24,7 +24,7 @@ env:
   APPLY_FIXES_EVENT: pull_request
   APPLY_FIXES_MODE: commit
 concurrency:
-  group: ${{ github.ref }}-${{ github.workflow }}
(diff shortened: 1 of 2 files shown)


MDr164:
I was already wondering where this behavior was coming from, nice that you already found that out

MDr164 (approved):

AtomicFS:
> I was already wondering where this behavior was coming from, nice that you already found that out

Yeah, I was thinking what is causing it and went down the rabbit-hole about `cancel-in-progress` and that it can take expressions, so I was thinking how to cancel only if the new job was not skipped ... and then it hit me - let's just change the `concurrency.group`.

PR #297, opened 2024-07-31
feat(release): automatically determinate the next version

AtomicFS:
- since we are using Conventional Commits, let's put them to work
- we can use the Conventional Commits to determine what the next version should be

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -33,26 +33,34 @@ jobs:
           fetch-depth: 0
           fetch-tags: true
 
-      - name: Get cocogitto
-        run: |
-          cargo install --locked cocogitto
+      # Automatically determinate the next release version from Conventional Commits since the latest tag
+      - name: Get next version
+        id: semver
+        uses: ietf-tools/semver-action@v1
+        with:
+          token: ${{ github.token }}
+          branch: main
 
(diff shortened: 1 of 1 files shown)


MDr164:
Will the action summarize the commits and just do a single bump from e.g. v1.2.0 to v1.3.0 or will it take each commit occurrence and add them together and eventually skip versions? I've seen some automated scripts do the later and I feel like that's incorrect. Do you already know or should I did through the actions code and figure it out?

AtomicFS:
> Will the action summarize the commits and just do a single bump from e.g. v1.2.0 to v1.3.0 or will it take each commit occurrence and add them together and eventually skip versions? I've seen some automated scripts do the later and I feel like that's incorrect. Do you already know or should I did through the actions code and figure it out?

Interesting question, did not occur to me. I was more interested if it requires `BREAKING CHANGE` string in commit body or if `feat!:` would be sufficient ([TLDR](https://github.com/ietf-tools/semver-action/issues/30): `!` is recognized).

But I was looking thought the code and it seems to do it in single increment.

AtomicFS:
Somebody actually wanted the behavior you describe in [issue 21](https://github.com/ietf-tools/semver-action/issues/21)

MDr164 (approved):

PR #304, opened 2024-08-02
feat(ci): add stale bot

AtomicFS:
- it will mark issues and PRs without activity as stale

### DIFF
--- a/.github/workflows/bot--stale.yml
+++ b/.github/workflows/bot--stale.yml
@@ -1,0 +1,29 @@
+---
+# Docs: https://github.com/actions/stale
+name: 'stale-issues-and-pull-requests'
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


MDr164:
I highly dislike the usage of 'stale' or even worse 'auto-close' just because something was sitting around for too long. Here is a better write-up of why I disagree with this feature: https://drewdevault.com/2021/10/26/stalebot.html

Proof me wrong with some actual benefits this gives us please.

AtomicFS:
> I highly dislike the usage of 'stale' or even worse 'auto-close' just because something was sitting around for too long. Here is a better write-up of why I disagree with this feature: https://drewdevault.com/2021/10/26/stalebot.html
> 
> Proof me wrong with some actual benefits this gives us please.

Oh I hate the idea of `auto-close` too, that is why I configured it to `days-before-close: -1` (effectively disabling this feature).

In this state the only thing it does is add label and a comment - basically just drawing attention to it.

I wholeheartedly agree with mister drewdevault here. As such there are no other benefits than having another auto-labeling bot.

AtomicFS:
Since you highly dislike the idea and I am not adamant about having it, let's just forget about it.

PR #314, opened 2024-08-22
feat(docker): build coreboot cross-compilers for all platforms

AtomicFS:
- if I am not mistaken we selected only few platforms to to save on compile time when building the docker containers
- since we are no longer building docker containers for each and every commit and/or pull request, I think we can support all platforms
- instead of 40 minutes per container it takes 60 minutes


AtomicFS:
@LeanSheng

AtomicFS:
@MDr164 review please

MDr164 (approved):
Yes, the main reason was compile time and container size but those are not that relevant as you noted due to the changes we did in the pipelines

PR #316, opened 2024-08-27
fix(action/linux): defconfig filename

AtomicFS:
We have discovered that the defconfig filename must end with `defconfig`.


MDr164 (approved):
Interesting that this is needed in their make system...

AtomicFS:
I hope it is also not needed for coreboot :D

PR #325, opened 2024-09-10
fix(lint): ignore review.coreboot.org links

AtomicFS:
The server is down for a while now and it prevents PR from merging


AtomicFS:
Seems to be up and running again.

PR #327, opened 2024-09-11
Fix/taskfile

### LINKED ISSUE #324: GOARCH is hardcoded to amd64 - builds firmware-action binary that doesn't run on arm64/aarch64
Building firmware-action on an arm64/AArch64 machine generates a binary that doesn't run, since it builds for amd64.
Removing `GOARCH` from Taskfile.yml fixes it.

AtomicFS:
- remove leftover junk in go-test workflow
- remove deploy-binaries task from taskfile, we no longer need to build all binaries with taskfile (binaries for releases are build with GoReleaser in release workflow)
- remove upx compression from taskfile since it just adds additional time with minimal benefits

fixes #324

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -8,7 +8,7 @@ tasks:
     desc: Template task to build a go binary
     dir: action
     cmds:
-      - go build -ldflags="-s -w" -o ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}}
+      - go build -ldflags="-s -w" -o ../bin/firmware-action-{{OS}}-{{ARCH}}-{{.SEMVER}}
     env:
       CGO_ENABLED: 0
 
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -67,9 +67,6 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: stable
-      - name: Install Taskfile
-        run: |
-          sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d
       - name: Install go dependencies
         run: |
           cd action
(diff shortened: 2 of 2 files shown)


Taskfile.yml
```diff
@@ -8,23 +8,10 @@ tasks:
     desc: Template task to build a go binary
     dir: action
     cmds:
-      - GOOS={{.OS}} GOARCH=amd64 go build -ldflags="-s -w" -o ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}}
-      - upx -9 ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}} || true
-      # upx works, but always fails on darwin:
-      #   CantPackException: macOS is currently not supported (try --force-macos)
+      - go build -ldflags="-s -w" -o ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}}
```
MDr164, line 11: I think the name here will no always hold true if the GOARCH is usersupplied.

Taskfile.yml
```diff
@@ -8,23 +8,10 @@ tasks:
     desc: Template task to build a go binary
     dir: action
     cmds:
-      - GOOS={{.OS}} GOARCH=amd64 go build -ldflags="-s -w" -o ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}}
-      - upx -9 ../bin/firmware-action-{{.OS}}-amd64-{{.SEMVER}} || true
-      # upx works, but always fails on darwin:
-      #   CantPackException: macOS is currently not supported (try --force-macos)
+      - go build -ldflags="-s -w" -o ../bin/firmware-action-{{OS}}-{{ARCH}}-{{.SEMVER}}
```
MDr164, line 11: While would always create a correct file name when compiling for the host target it might still create confusion when cross compiling in Go, given that `OS` and `Arch` are Taskfile specific variables that reflect the host calling the Taskfile.

AtomicFS: Yes, but I removed any option for cross-compilation from the taskfile. So that scenario should not happen.

I could go the other way around, keep cross-compilation, make this task `internal` and parametric and create instead tasks like `build-go-binary-linux` or `build-go-binary-windows` and so on.

PR #332, opened 2024-09-17
feat(ci): use arduino/setup-task to get taskfile

AtomicFS:
Must be merged after #330


MDr164 (approved):
Oh yeah, I forgot that action exists. Will probably equivalent to the curl execution but I'm fine with either.

PR #333, opened 2024-09-17
Feat/edk2 202408

AtomicFS:
Must be merged after #332 

So I tried to add new `edk2-stable202408` container and oh boy, there was a lot to do. Pardon me but I will be a bit verbose here (mostly for sake of future me):
- I had some issues during building the containers locally, so I had to debug the Python code that have not been touched in a long time. Thankfully it was just `dagger` issue, so I added more helpful error messages. And while doing that I also removed some old Python code that seemed to be useless.
- Then I ran out of disk space while building the container (yet again) so I had to further slim down the containers.
  - I have done that by applying the same trick I used for `edk2-platforms` and `edk2-non-osi` repositories to `edk2`, but this required changes in the `Dockerfile`s and `compose.yaml`.
  - I was firmly reminded why I did not do it the first time around - the ancient `UDK2017` was a problematic one. There is a tag `UDK2017` and branch `vUDK2017` and they are completely different. And of course I did accidentally (again) switch from functional `UDK2017` branch to broken `vUDK2017` tag by using wrong commit hash in `compose.yaml`. This time I noticed what was the actual problem which allowed me to use this disk-saving approach.
  - There is still a problem of too much disk space being used, so I temporarily disabled testing of the docker containers - this should not be a problem since our containers have been rock-solid for quite some time now.
  - This problem should still be addressed (must be addressed before next release), so I opened a issue #338 
- As if this was not enough, I had issues with building the newest `edk2` container, while all previous ones worked fine. I was informed that this is caused by recent change that requires `-D BUILD_ARCH=X64` to be added to the build command. Thankfully this does not cause problems in old containers.
- Because it has been a while since I added new container, I kinda forgot all the steps, so I added notes on this topic into the documentation.

### DIFF
--- a/.dagger-ci/daggerci/lib/orchestrator.py
+++ b/.dagger-ci/daggerci/lib/orchestrator.py
@@ -204,9 +204,7 @@ class Orchestrator:
                             dockerfile,
                         )
                 else:
-                    await self.__build_test_publish__(
-                        client, top_element=top_element, dockerfile=dockerfile
-                    )
+                    await self.__build_test_publish__(client, top_element, dockerfile)
 
         return self.results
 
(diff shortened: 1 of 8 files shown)


AtomicFS:
@MDr164 this should work now. Review pls.

MDr164 (approved):
Your PR text explains well enough why those TODO changes had been necessary so I'd say this is good to go under the current circumstances.

PR #339, opened 2024-09-24
chore(python): switch from hardcoded version to latest stable

AtomicFS:


MDr164 (commented):
Does python guarantee no sudden breakage in every 3.x release?

AtomicFS:
There are no guarantees, but it should work fine. I am just paranoid that the version will get forgotten and stuck at some point.

It already happened in `github/workflows/pytest.yml` file which was not udated to python `v3.12`.

MDr164 (approved):
Let's see if this works better from a maintenance point of view

PR #343, opened 2024-09-25
feat(docker): make containers multi-arch

### LINKED ISSUE #341: add multi-architecture support (arm64)
It would be nice to add support for other architectures other than `x86`.

The golang code is already compiled for `arm64`, `i386` and `x86_64`. And `Docker` containers do support [multi-arch images](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/). Not only that, dagger which we are using [supports this too](https://docs.dagger.io/cookbook/#build-multi-arch-image).

- [x] Add multi-arch support to python code responsible for building the containers
- [x] Add multi-arch support to golang code
  - ~~this might already be supported by default because of docker / dagger~~
  - this is handled by dagger / docker when we provide multi-platform containers

AtomicFS:
This makes all of our containers [multi-architectural / multi-platform](https://docs.docker.com/build/building/multi-platform/) and fixes #341 .

At the moment we only support `x86_64` (aka `amd64`) and `arm64` (specifically `arm64 v8`).

Todo:
- [x] uroot (this was no work)
- [x] linux (this was easy)
- [x] coreboot (took some effort but works)
- [x] edk2 (this is pain)
  - [x] build the toolchains on the fly for non-x86 architectures (first use by user)
    - this is kinda difficult to test at the moment

There is a lot to comprehend, so here are some notes as to why:
- `u-root` and `linux` are OK to build the "normal way", however with `edk2` and `coreboot` I get errors when trying to compile to tool-chains - I suspect the problem to be native emulation not supporting all of the instructions required
  - building `x86` on `x86` machine and `arm64` on `arm64` works fine, but building `arm64` on `x86` machine is broken (and vice versa)
- this means that I cannot build `edk2` or `coreboot` multi-arch container on single `x86` machine as I can `u-root` or `linux`
- I looked into cross-compilation - aka cross-compile the cross-compilation tool-chains ... well that did not go well
  - it should be possible, but it seems like a long process not worth the time investment
- one way would be to build each arch-container separately (`x86` part on `x86` machine and `arm64` part on `arm64` machine), publish them separately and then publish hand-crafted manifest connecting them together ... (see `The hard way with docker manifest` in [Multi-arch build and images, the simple way](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/)
- but we can avoid all of that with simply compiling the tool-chains separately on their respective machines and then just copy files into single multi-arch container (which can be done without any problems with simple native emulation)
  - **UPDATE**: `coreboot` is fine, but `edk2` is too much hassle (see #368)
- one advantage of separate tool-chain compilation is that this way we can cache the compiled tool-chains for `coreboot` and speed up the build by one order of magnitude (normally they take around 1 hour to compile)
- as I was running out of disk-space for coreboot docker container I decided to use [upx](https://archlinux.org/packages/extra/x86_64/upx/) to compress the toolchain reducing the size from around 1.5 GB down to 1.0 GB (as a nice side-effect the whole container is smaller)
- as it stands now, we are not testing if the multi-arch containers work on other than `x86` architectures, since it is not really possible right now. We can only verify that the new changes did not break existing functionality.

This is the final PR in the multi-arch saga, which contains breaking changes.
- #355
- #356
- #359
- #360
- #361
- #362
- #366
- #367
- #368

### DIFF
--- a/.github/workflows/docker-build-and-test.yml
+++ b/.github/workflows/docker-build-and-test.yml
@@ -255,7 +255,7 @@ jobs:
   build:
     name: build_test_publish
     runs-on: ubuntu-latest
-    timeout-minutes: 120
+    timeout-minutes: 60
     needs:
       - get-matrix
       - build-coreboot-toolchains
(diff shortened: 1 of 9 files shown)


AtomicFS:
The linting issue is a [known bug](https://github.com/hadolint/hadolint/issues/951)

.github/workflows/docker-build-and-test.yml
```diff
@@ -302,6 +302,95 @@ jobs:
...
+        if: startsWith(matrix.dockerfile, 'coreboot')
+        run: |
+          git clone --depth 1 "https://review.coreboot.org/coreboot.git" -b "${{ steps.version.outputs.COREBOOT_VERSION }}"
+      - name: Get coreboot commit hash
+        id: coreboot-hash
+        if: startsWith(matrix.dockerfile, 'coreboot')
+        run: |
+          cd coreboot
+          COREBOOT_HASH="$( git rev-parse --short HEAD )"
+          echo "${COREBOOT_HASH}"
```
MDr164, line 319: What is the purpose of outputting the content of the variable here? Just for debugging? And in the next step you out `CACHE_KEY=${CACHE_KEY}` aka `KEY=VALUE` and here just `VALUE`. Was that done on purpose?

AtomicFS: Just for debugging.

MDr164 (approved):
I hope I didn't miss anything but looks pretty solid to me

AtomicFS:
I could then add tests for `arm64` as well.

PR #346, opened 2024-09-26
fix(docker): enable again tests of containers

### LINKED ISSUE #338: building docker containers uses too much disk space
This is causing problems for the latest `edk2-stable202408` container.

Main problem is that the container is exported as tarball, and then imported back to being a testing (execution of test script). `edk2` containers take up around 2.1 GB, with export and import that means 3 x 2.1 GB (6.3 GB) which overflows GitHub's 14 GB disk space.

This whole export and import process is unnecessary. We should switch to dagger way of branching off a separate container for testing (just like we do for interactive mode).

AtomicFS:
This time the tests are done without tarball export and import which should save up a lot of disk space and a bit of CI time.

fixes #338

### DIFF
--- a/.dagger-ci/daggerci/lib/orchestrator.py
+++ b/.dagger-ci/daggerci/lib/orchestrator.py
@@ -12,13 +12,6 @@ import re
 import sys
 from typing import Any
 
-try:
-    import humanize
-except ImportError:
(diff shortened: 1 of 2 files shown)


MDr164 (approved):
CI seems happy so far :+1:

PR #349, opened 2024-09-26
fix(edk2): GCC5 was depreciated in edk2-stable202305

### LINKED ISSUE #340: edk2: change `GCC5` to `GCC` since `edk2-stable202305`
Edk2 toolchain definitions was updated to deprecate `GCC5` since newer `gcc` versions work as well.

In multiple locations we are still using `GCC5`, instead we should use `GCC`.

For example `tests/test_edk2.sh`
```patch
- build -D BOOTLOADER=COREBOOT -a IA32 -a X64 -t GCC5 -b DEBUG -p "${PAYLOAD}"
+ build -D BOOTLOADER=COREBOOT -a IA32 -a X64 -t GCC -b DEBUG -p "${PAYLOAD}"
```

Then there are also:
- `tests/example_config.json`
- `action/recipes/edk2_test.go`
- `action/recipes/edk2.go`

For more details see commit [BaseTools/Conf/tools_def.template: Add section for deprecated toolchains](https://github.com/tianocore/edk2/commit/0fc07b1c6a491fa1e81daed6cfc2ec33c8cac973)

AtomicFS:
fixes #340

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -121,6 +121,7 @@ jobs:
       - name: Move my defconfig into place (filename must not contain '.defconfig')
         run: |
           mv "tests/coreboot_${{ matrix.coreboot-version }}/seabios.defconfig" "seabios_defconfig"
+
       - name: firmware-action
         uses: ./
         # uses: 9elements/firmware-action
@@ -131,6 +132,7 @@ jobs:
           compile: ${{ needs.changes.outputs.action }}
         env:
           COREBOOT_VERSION: ${{ matrix.coreboot-version }}
+
       - name: Get artifacts
         uses: actions/upload-artifact@v4
         with:
@@ -184,6 +186,7 @@ jobs:
       - name: Move my defconfig into place (filename must not contain '.defconfig')
         run: |
           mv "tests/linux_${{ matrix.linux-version }}/linux.defconfig" "ci_defconfig"
+
       - name: firmware-action
         uses: ./
         # uses: 9elements/firmware-action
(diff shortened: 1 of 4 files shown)


.github/workflows/example.yml
```diff
@@ -237,6 +241,23 @@ jobs:
...
 
+      - name: Get versions of edk2
+        id: edk2_versions
+        run: |
+          echo "ver_current=$( echo ${{ matrix.edk2-version }} | tr -cd '0-9' )" >> "${GITHUB_OUTPUT}"
+          echo "ver_breaking=$( echo 'edk2-stable202305' | tr -cd '0-9' )" >> "${GITHUB_OUTPUT}"
+
+      - name: Use GCC5 for old edk2
+        id: gcc_toolchain
+        # GCC5 is depreciated since edk2-stable202305
```
MDr164, line 252: Just a nit: I never saw 'depreciated' in a programming context so I wondered if it means the same as 'deprecated'. According to this it does not: https://stackoverflow.com/a/9208164 

Given that I'm not a native English speaker I have no real preference here but it just seemed odd.

tests/test_edk2.sh
```diff
@@ -54,5 +54,14 @@ fi
 if [ "${VERIFICATION_TEST_EDK2_VERSION}" == "UDK2017" ]; then
 	OvmfPkg/build.sh -a X64
 else
-	build -D BOOTLOADER=COREBOOT -a IA32 -a X64 -t GCC5 -b DEBUG -p "${PAYLOAD}" -D BUILD_ARCH=X64
+	# GCC5 is depreciated since edk2-stable202305
```
MDr164, line 57: Same here

.github/workflows/example.yml
AtomicFS: Oh yeah :D
I did not relalise that those are 2 different terms xD
I assumed it is like `color` vs `colour`.

MDr164 (approved):

PR #356, opened 2024-10-10
ci(docker): build monthly instead of weekly

AtomicFS:
- I don't think we need to re-build the containers this often


MDr164 (approved):
I agree, we just don't have the change frequency that would justify weekly

PR #361, opened 2024-10-10
feat: dockerfile support multi arch

AtomicFS:
Preparing for multi-arch support, `arm64` specifically.

I did test the functionality manually on `arm64` machine, so I guess it is fine for now. I definitely still works for `x86` (`amd64`) since the tests are passing.

If there is some issue with `arm64` builds, then I can fix them in PR which will actually use this feature.

### DIFF
--- a/docker/coreboot/Dockerfile
+++ b/docker/coreboot/Dockerfile
@@ -1,12 +1,15 @@
 # This is multi-stage Dockerfile
 #   Docs: https://docs.docker.com/build/building/multi-stage/
 
+ARG TARGETARCH=amd64
+
 #=============
 # "base" stage with all needed build dependencies
 ARG SOURCE_IMAGE=ubuntu:jammy
 FROM ${SOURCE_IMAGE} AS base
 
-ARG TARGETARCH=amd64
+ARG TARGETARCH
+
 ARG COREBOOT_VERSION=4.19
 ARG CONTEXT=coreboot
 
@@ -32,7 +35,6 @@ RUN apt-get update && \
         git \
         gnat \
         imagemagick \
-        iucode-tool \
         libelf-dev \
         libncurses5-dev \
         libnss3-dev \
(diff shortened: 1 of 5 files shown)


docker/coreboot/Dockerfile
```diff
@@ -1,11 +1,14 @@
...
 #   Docs: https://docs.docker.com/build/building/multi-stage/
 
+ARG TARGETARCH=amd64
+
 #=============
 # "base" stage with all needed build dependencies
 FROM ubuntu:jammy AS base
 
-ARG TARGETARCH=amd64
+ARG TARGETARCH
```
MDr164, line 10: Can you briefly explain why we define TARGETARCH twice, once with a value (like previously) and now a second time without one?

docker/linux/Dockerfile
```diff
@@ -52,6 +49,12 @@ RUN apt-get update && \
         wget \
         zstd \
     && \
+    if [ "${TARGETARCH}" = 'amd64' ]; then \
```
MDr164, line 52: Shouldn't we still install different GCC compilers on non amd64 as well? E.g. if one uses arm64 natively but wants to cross compile for amd64 instead.

docker/coreboot/Dockerfile
AtomicFS: Oh yeah, sorry.

It is because of the `ARG` limited [scope](https://docs.docker.com/reference/dockerfile/#scope).

> An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an argument in multiple stages, each stage must include the ARG instruction.

TLDR: `FROM` starts a new scope (a new stage) where the previous `ARG` does not apply anymore. It is possible to use `ARG` without value if the previous `ARG` was declared in the top scope (aka before first `FROM`).

docker/linux/Dockerfile
AtomicFS: Yes, but I think I included this because I had issues ... but that is a lot of changes and bug-fixes ago, so I will test it again.

AtomicFS: Jup, seems to actually work fine. I will remove the `IF` and cleanup it up a bit.

I also noticed that I by accident included the `test_linux.sh` file, I will remove that for now.

AtomicFS: OK, this should be much better now.

MDr164 (approved):

PR #362, opened 2024-10-10
ci(docker): dynamically generate the matrix

AtomicFS:
- instead of hard-coded matrix of docker containers, generate it automatically
- I also include matrix subsets only with `coreboot` and `edk2` entries, they will be needed later

### DIFF
--- a/.github/workflows/docker-build-and-test.yml
+++ b/.github/workflows/docker-build-and-test.yml
@@ -30,36 +30,61 @@ permissions:
   packages: write
 
 jobs:
+  #=============================
+  # Dynamically generate matrix
+  #=============================
+
+  get-matrix:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: get-matrix
+        id: get-matrix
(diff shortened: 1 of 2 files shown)


AtomicFS:
@MDr164 review pls?

.github/workflows/docker-build-and-test.yml
```diff
@@ -30,36 +30,61 @@ permissions:
...
+  get-matrix:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: get-matrix
+        id: get-matrix
+        run: |
+          # Disable SC2046: Quote this to prevent word splitting
+          # I can't quite this, just look at it
```
MDr164, line 46: I think there is either a word missing here or the `e` is too much in quite

MDr164 (commented):
Looks good so far, just a minor error in a comment

.github/workflows/docker-build-and-test.yml
AtomicFS: Typo :)

PR #368, opened 2024-10-14
feat(docker): compile toolchains separately

AtomicFS:
At the moment this is not used yet, but will be.

Basically I am moving the compilation of the tool-chains out of the Dockerfile, as I believe this to be the only reasonable way to achieve multi-arch docker containers.

There is a lot to comprehend, so here are some notes as to why:
- `u-root` and `linux` are OK to build the "normal way", however with `edk2` and `coreboot` I get errors when trying to compile to tool-chains - I suspect the problem to be native emulation not supporting all of the instructions required
  - building `x86` on `x86` machine and `arm64` on `arm64` works fine, but building `arm64` on `x86` machine is broken
- this means that I cannot build `edk2` or `coreboot` multi-arch container on single `x86` machine as I can `u-root` or `linux`
- I looked into cross-compilation - aka cross-compile the cross-compilation tool-chains ... well that did not go well
- one way would be to build each arch-container separately (`x86` part on `x86` machine and `arm64` part on `arm64` machine), publish them separately and then publish hand-crafted manifest connecting them together ... (see `The hard way with docker manifest` in [Multi-arch build and images, the simple way](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/)
- but we can avoid all of that with simply compiling the tool-chains separately on their respective machines and then just copy files into single multi-arch container (which can be done without any problems with simple native emulation)
- so this is one step towards that
- one advantage is that this way we can cache the compiled tool-chains for `coreboot` and speed up the build by one order of magnitude (normally they take around 1 hour to compile)

**UPDATE**:
- unfortunately pre-compiling `edk2` `BaseTools` is not that simple, there are 2 big problems:
  - the oldest `edk2` that we have is `udk2017` which is using `Ubuntu Bionic` (`18.04`) which is simply too old to run GitHub action inside it due to obsolete `NodeJS` (see [GLIBC_2.28 not found](https://github.com/actions/checkout/issues/1590))
  - GitHub is not going to fix this, nor they are likely to remove the idiotic design-flaw of having actions being written in JavaScript
- so I will make it such that the `edk2` `BaseTools` gets compile on the first use by the user, and since it does not take long, it should be fine I guess


AtomicFS:
@marvwhere I think that this is ready for preview.

Next PR will replace the in `Dockerfile` toolchain compilation with these pre-compiled toolchains.

.github/workflows/docker-build-and-test.yml
```diff
@@ -58,33 +60,212 @@ jobs:
...
+            ca-certificates \
+            curl \
+            git \
+            jq \
+            sudo \
+            tzdata \
+            upx-ucl \
+            wget
+          update-ca-certificates
+      - name: Get yq
```
MDr164, line 105: I'm a bit confused here. So you fetch `yq` in this step which makes it available for this jobs following steps but inside the get-matrix(-coreboot) jobs you also call into `yq` without fetching it first, implying that it was already present. Was it not present on jammy or am I missing something here?

MDr164 (commented):
Looks alright, gonna run some local tests once those updates are in

.github/workflows/docker-build-and-test.yml
AtomicFS: huh, that is a good question ...

AtomicFS: According to [github runner-images](https://github.com/actions/runner-images?tab=readme-ov-file#available-images) `ubuntu-latest` == `22.04` == `jammy`.

So technically speaking everything runs on `jammy` ...

MDr164: So I guess fetching yq is not needed then or do the two matrix jobs not work right now?

AtomicFS: Oh yeah, there is a difference. `runs-on:` is usign GitHub's flavoured Docker containers where `yq` [is installed](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools), but `container:` is [using upstream Docker-hub images](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses) which are minimal.

AtomicFS: `get-matrix` and `get-matrix-coreboot` are using `runs-on: ubuntu-latest` which has `yq` installed, but `build-coreboot-toolchains` runs on various GitHubs ubuntu containers (based on config in `docker/compose.yaml` it is only `jammy` at the moment) but then they spawn a Docker container `container: image: ubuntu:jammy` which is from Docker-hub without `yq`.

MDr164: I see, thanks for clarifying!

MDr164 (approved):

PR #371, opened 2024-10-22
docs: add SECURITY.md

AtomicFS:
Not sure if we should use this email address, might end up on some spam-bot list. But it is already listed on our website.

I also do not want to use email address for any individual to mitigate [the bus factor](https://en.wikipedia.org/wiki/Bus_factor).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,18 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


MDr164 (approved):
That mail is not used for outgoing communication afaik so should be fine to have it listed here as well

PR #373, opened 2024-10-25
fix(docker): typo in compose.yaml

AtomicFS:
Not critical but, but it bugs me :D (pun intended)


MDr164 (approved):
The OCD was hitting on that one :wink:

PR #377, opened 2024-10-30
expand taskfile

AtomicFS:
I am adding a new task into the `Taskfile.yml` inside `.dagger-ci` directory, which simplifies building single docker containers (mostly for development purposes).

I am also linking the the two taskfiles (the main one and one in the `.dagger-ci` directory) for simplicity.

Building coreboot containers will be a bit problematic after the #343 is merged because it will require to have pre-compiled tool-chain ready on disk :(

### DIFF
--- a/.dagger-ci/Taskfile.yml
+++ b/.dagger-ci/Taskfile.yml
@@ -2,6 +2,7 @@
 version: '3'
 vars:
   PROJECT: 'daggerci'
+  PYTHON_VENV: 'venvdagger'
 
 tasks:
   lint-pylint:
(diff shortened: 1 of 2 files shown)


.dagger-ci/Taskfile.yml
```diff
@@ -27,3 +28,70 @@ tasks:
     dir: daggerci
     cmds:
       - python -m pytest --cov --cov-report=term-missing --cov-report=html --log-cli-level NOTSET --show-capture no --log-cli-level=INFO --runslow {{.CLI_ARGS}}
+
+  setup-python-venv:
+    desc: Setup python virtual environment
+    internal: true
+    cmds:
+      - python -m venv {{.PYTHON_VENV}}
+      - bash -c "
```
MDr164, line 37: Are those multiline cmds working without `\` at the end or prefixing it with the yaml `|` syntax? Looks like GitHubs syntax highlighter does not like it

AtomicFS: It does seem to work ... works on my machine xD

AtomicFS: Yeah, I added the `|` character. GitHubs syntax highlighter should be happier now :)

PR #388, opened 2024-11-06
test(examples): expand to also run on arm64 machine

AtomicFS:
- add example test for `coreboot` on `arm64`
- add example test for `uroot` on `arm64`
- add example test for `linux` on `arm64`
- add example test for `stitching` on `arm64`

edk2 will come in later PR

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -145,7 +145,7 @@ jobs:
       - name: Get artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: coreboot-${{ matrix.coreboot-version }}
+          name: coreboot-${{ matrix.coreboot-version }}-${{ matrix.arch }}
           path: output-coreboot
           retention-days: 14
   # ANCHOR_END: example_build_coreboot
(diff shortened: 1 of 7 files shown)


AtomicFS:
The issue is failing because of missing git submodule, which is fixed in #392 , so should be fine too @MDr164

AtomicFS:
There was some left-over garbage in `arm64` runner which was causing problems. Is fixed now.

.github/workflows/example.yml
```diff
@@ -89,16 +89,22 @@ jobs:
   # Example of building coreboot
   # ANCHOR: example_build_coreboot
   build-coreboot:
-    runs-on: ubuntu-latest
```
MDr164, line 92: Could you briefly explain why this got removed? Otherwise the PR looks fine to me.

AtomicFS: Yes, it is replaced with
```yaml
runs-on: ${{ matrix.arch == 'arm64' && 'ARM64' || 'ubuntu-latest' }}
```

So basically it picks whether to run on `ubuntu-latest` or our self-hosted runner.

MDr164: Oh I missed that line, thanks for the explanation :+1:

MDr164 (dismissed):

AtomicFS:
Ups, the artifacts had conflicting names. Fixed.

MDr164 (approved):

PR #392, opened 2024-11-07
fix(docker): edk2-stable202408 missing submodule

AtomicFS:
replace edk2-stable202408 with edk2-stable202408.01 to fix the missing git submodule

fixes #385

PR #394, opened 2024-11-08
fix(dagger): install build-essential package

AtomicFS:


MDr164 (approved):
I'm surprised this package wasn't already installed

AtomicFS:
Should be, but I got some random out-of-nowhere error that `make` is not found in #392

PR #402, opened 2024-11-12
ci(lint): depreciate commitlint in favour of bugbundle/commits

AtomicFS:


AtomicFS:
This looks like more work than I am willing to put in

PR #406, opened 2024-11-14
feat(docker)!: bump GCC_VERSION for edk2-202408.01

### LINKED ISSUE #413: Example tests download firmware-action from releases instead of compilation
Oh, I know now why it behaves the way it does. In examples it is not compiling the `firmware-action` binary, it is fetching it from releases on GitHub. But we had some breaking changes since.

Problem is that the example CI does not detect that `firmware-action` needs to be compiled in this instance. It detects if compilation is needed based on diff of `action/` directory compared to `main` and not compared to latest release.

_Originally posted by @AtomicFS in https://github.com/9elements/firmware-action/issues/410#issuecomment-2477265624_

AtomicFS:
@LeanSheng I hope this is correct

among other things also fixes #413

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -73,19 +73,35 @@ jobs:
     permissions:
       pull-requests: read
     outputs:
-      action: ${{ steps.filter.outputs.action }}
+      compile: ${{ steps.compile.outputs.compile }}
     if: ${{ needs.skip-check.outputs.changes == 'true' }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: dorny/paths-filter@v3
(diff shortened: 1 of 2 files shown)


AtomicFS:
Once the #409 is merged and CI clears up a bit I will rebase this and bump it to GCC 13 to see if it works.

AtomicFS:
edk2 docker builds are failing because of:
```
fatal: unable to access 'https://boringssl.googlesource.com/boringssl/': The requested URL returned error: 503
```

AtomicFS:
@MDr164 finally ready for review

AtomicFS:
> Let's see if GCC13 works well enough for this version of EDK2

Well, the container self-test does work. So let'f hope that the test is working as intended.

PR #409, opened 2024-11-14
fix(examples): partial revert of 32583f79

AtomicFS:
- in action.yml the ARCH change is broken since 32583f79
- for more details see comment in the file


MDr164 (approved):
stamped

PR #410, opened 2024-11-14
fix(examples): fixed typo

AtomicFS:


MDr164 (approved):

AtomicFS:
@MDr164 Works now without changing a thing ... only added a debug code ... don't know what is going on, probably some GitHub nonsense.

MDr164 (approved):
GitHub being weird then

AtomicFS:
Oh, I know now why it behaves the way it does. In examples it is not compiling the `firmware-action` binary, it is fetching it from releases on GitHub. But we had some breaking changes since.

Problem is that the example CI does not detect that `firmware-action` needs to be compiled in this instance. It detects if compilation is needed based on diff of `action/` directory compared to `main` and not compared to latest release.

PR #412, opened 2024-11-14
test(examples): add taskfile to run Linux example locally

AtomicFS:
Comes handy for development and debugging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .coverage
 .task
+.example_runs
 action/coverage.html
 action/coverage.out
 bin/
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -8,6 +8,11 @@ includes:
     taskfile: ./.dagger-ci/Taskfile.yml
     dir: ./.dagger-ci
     optional: true
+  examples:
+    taskfile: ./tests/Taskfile.yml
+    optional: true
+    vars:
+      SEMVER: '{{.SEMVER}}'
 
 tasks:
   build-go-binary:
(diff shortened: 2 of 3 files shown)


AtomicFS:
> How do you actually invoke this task? I just tried calling it like so: `task examples:linux-*` but got `template: :1:2: executing "" at <index .MATCH 0>: error calling index: index of untyped nil` as an error then.

Taskfile has a [wildcard arguments](https://taskfile.dev/usage/#wildcard-arguments) feature, which is what I used there.

TLDR: Do not run `task examples:linux-*` but `task examples:linux-6.9.9`

MDr164:
Thanks, that allows running the task but I'm getting all kinds of different errors depending on what I'm doing. For example choosing a version at random will error out as there is no defconfig (maybe we should have a wildcard one) and when choosing one we got a config for I get `firmware-action/bin/firmware-action-linux-amd64-v0.6.1: no such file or directory` when running from a clean repo. I guess the task assumes some existing files that are not part of deps or sources/generates.

AtomicFS:
It is likely it fails because of missing files. I did not tests that, just made it run from my messy repository :D

I will look into it and fix it :)

AtomicFS:
Oh yeah, found the problem. The example does not run because the firmware-action executable does not exist by default.

Thankfully I can easily fix this by adding dependency of this included task on task in root taskfile.

> NOTE: If you want to call a task declared in the root Taskfile from within an [included Taskfile](https://taskfile.dev/usage/#including-other-taskfiles), add a leading `:` like this: `task: :task-name`.

Source: [Docs - look at the TIP](https://taskfile.dev/usage/#calling-another-task)

AtomicFS:
Fixed

MDr164 (approved):
Works!

PR #414, opened 2024-11-15
Ci/store coreboot toolchains in new repository

AtomicFS:
Alright, here is some fun stuff.

With this change, we will use another GitHub repository [firmware-action-toolchains](https://github.com/9elements/firmware-action-toolchains/) in place of cache for compiled coreboot toolchains. When we add new coreboot version, it will automatically commit and push the new toolchain into the remote repository and create a pull request.

Caching is still enabled to speed-up development (mostly development of this new flow).

The toolchains are now exclusively downloaded from the remote repository [firmware-action-toolchains](https://github.com/9elements/firmware-action-toolchains/) for docker building. So when a new coreboot version is added, the CI with that new version will very likely fail because the toolchain will not be yet merged. Simple re-run will remedy this issue.

### DIFF
--- a/.github/workflows/docker-build-and-test.yml
+++ b/.github/workflows/docker-build-and-test.yml
@@ -342,56 +342,32 @@ jobs:
           COREBOOT_HASH="$( git rev-parse --short HEAD )"
           echo "${COREBOOT_HASH}"
           echo "COREBOOT_HASH=${COREBOOT_HASH}" >> "${GITHUB_OUTPUT}"
-      - name: Artefact and cache key
-        id: cache-key
-        if: startsWith(matrix.dockerfile, 'coreboot')
-        run: |
-          CACHE_KEY="coreboot-${{ steps.version.outputs.COREBOOT_VERSION }}-${{ steps.coreboot-hash.outputs.COREBOOT_HASH }}"
-          echo "CACHE_KEY=${CACHE_KEY}"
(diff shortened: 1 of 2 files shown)


AtomicFS:
The token needs to be approved

AtomicFS:
@MDr164 btw, this would really speed up the whole CI.

.github/workflows/docker-build-and-test.yml
```diff
@@ -185,69 +198,79 @@ jobs:
...
-          path: ${{ steps.cache-key.outputs.CACHE_KEY }}-xgcc.tar
-          retention-days: 30
-          include-hidden-files: true
-          compression-level: 9
-      - name: Upload utils
-        uses: actions/upload-artifact@v4.4.3
+          repository: '9elements/firmware-action-toolchains'
+          path: 'firmware-action-toolchains'
+          ref: 'main'
+          lfs: false
```
MDr164, line 246: I'm surprised this works without LFS

MDr164 (approved):
Nice! It feels slightly hacky in some places but all in all looks solid enough to solve our cache problems :+1:

.github/workflows/docker-build-and-test.yml
AtomicFS: The `actions/checkout` documentation:
```
    # Whether to download Git-LFS files
    # Default: false
    lfs: ''
```

PR #415, opened 2024-11-15
fix(ci/automerge): use PAT token

AtomicFS:
My theory as to why GitHub does not trigger check jobs when Auto-merge is enabled by our bot is that GitHub does not allow ordinary `GITHUB_TOKEN` to trigger other workflows. Probably to avoid endless recursion.

We had this problem also with releases. So I hope that using PAT might fix this.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -51,6 +51,7 @@
     "acpica",
     "addinivalue",
     "addoption",
+    "automerge",
     "autopep",
     "blkio",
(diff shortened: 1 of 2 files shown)


AtomicFS:
The token was not approved yet, so this is failing at the moment.

MDr164 (approved):

MDr164:
Seems like it's working with the PAT

PR #418, opened 2024-11-19
chore(dependabot): remove docker

AtomicFS:
- we never use it anyway
- when I added it I was young and foolish, did not read the documentation properly
- all it would do is to add some metadata, and now I do not see any benefits

[Docs](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#docker)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,17 +23,6 @@ updates:
           - 'minor'
           - 'patch'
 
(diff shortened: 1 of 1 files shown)


MDr164 (approved):
To be fair I also did not read the docs there and expected some more value but looks like we can remove it

PR #419, opened 2024-11-19
fix(action): linux make defconfig file conflict

AtomicFS:
- instead of copy use move on the defconfig


MDr164 (approved):
make is just wonderful

PR #421, opened 2024-11-21
feat(action): add shell completion

### LINKED ISSUE #188: Add `bash-completion`

AtomicFS:
Not sure what is the best way to do this, but since `kong` does not support shell completion out of the box, and I don't want to write all the shell completion scripts by hand, I tried to use [complgen](https://github.com/adaszko/complgen) and it seems to work really nicely.

So the `firmware-action.usage` file is the one I written by hand, and the `firmware-action.(bash|fish|zsh)` files are generated by `complgen`.

When we get this into release, I can update the AUR packages to put these scripts into proper places to work seamlessly.

You can test the completion script just by sourcing it and then trying to write `firmware-action` command.

fixes #188


MDr164 (commented):
confirmed working with zsh and bash but only if the binary is called "firmware-action" and not something like "firmware-action-linux-amd64-v0.6.1"

AtomicFS:
> confirmed working with zsh and bash but only if the binary is called "firmware-action" and not something like "firmware-action-linux-amd64-v0.6.1"

Isn't that how it should be?

MDr164:
I suppose so, I was just curious if there could be a wild card like "firmware-action*" but on a second thought that might also not be the smartest.

MDr164 (approved):

PR #447, opened 2024-11-28
Chore/move golang code

AtomicFS:


MDr164:
Does this work as expected btw? If I recall correctly the reason to change it was due to how `go install` finds and names the binary, right?

AtomicFS:
> Does this work as expected btw? If I recall correctly the reason to change it was due to how `go install` finds and names the binary, right?

Yes, that is the reason. It should work, but I do not know how to actually test it.

MDr164:
I'll try to test this out from the branch soon to check if this does what we want.

MDr164:
> I'll try to test this out from the branch soon to check if this does what we want.

```
❯ go install github.com/9elements/firmware-action/cmd/firmware-action@65d6577647f8224a70aa7f09fe4e199e603bb5fd
go: downloading github.com/9elements/firmware-action v0.10.3-0.20250107091606-65d6577647f8
go: downloading github.com/9elements/firmware-action/cmd/firmware-action v0.0.0-20250107091606-65d6577647f8
```

Looks like it is trying to fetch the right thing so should be fine to go in. I'll double check if everything is moved to the new path without any leftover references to the old one.

MDr164:
Needs one last rebase but should be good to go then

AtomicFS:
Do you want to do it?

MDr164 (dismissed):
Let's hope this does not break anything :crossed_fingers:

MDr164:
Need your approval as well @AtomicFS to merge

AtomicFS:
It says that "Pull request authors can’t approve their own pull request."

AtomicFS:
Just bypass the rules.

MDr164:
> It says that "Pull request authors can’t approve their own pull request."

Also says "New changes require approval from someone other than MDr164 because they were the last pusher." :D

AtomicFS:
Done

FYI, you also committed a compiled go executable, so I removed it ;)

MDr164 (approved):

PR #453, opened 2024-12-02
Docs/offline usage and troubleshooting

### LINKED ISSUE #448: firmware-action fails when running offline
When `firmware-action` is executed without internet connection, it fails to fetch the contianer (even though it is already present locally) and fails the whole build.
```
- error: input: container.from resolve: failed to resolve image ghcr.io/9elements/firmware-action/udk2017:v0.8.0: failed to resolve source metadata for ghcr.io/9elements/firmware-action/udk2017:v0.8.0: failed to do request: Head "https://ghcr.io/v2/9elements/firmware-action/udk2017/manifests/v0.8.0": dial tcp: lookup ghcr.io: i/o timeout
```

This error should not cause `firmware-action` to fail all together, at most it should give a warning and carry on if there is a container already present.

AtomicFS:
fixes #448

### DIFF
--- a/docs/src/firmware-action/offline_usage.md
+++ b/docs/src/firmware-action/offline_usage.md
@@ -1,0 +1,33 @@
+# Offline usage
+
+`firmware-action` under the hood uses [dagger](https://docs.dagger.io/) / docker. As such, the configuration contains entry `sdk_url` which specifies the docker image / container to use.
+
+```admonish example
+~~~json
+"sdk_url": "ghcr.io/9elements/firmware-action/coreboot_4.19:main"
+~~~
+```
+
+However, in this configuration `firmware-action` (or rather `dagger`) will always connect to the internet and download the manifest to see if a new container needs to be downloaded. This applies to all tags (`main`, `latest`, `v0.8.0`, and so on).
+
+If you need to use `firmware-action` offline, you have to first acquire the container. Either by running `firmware-action` at least once online, or by other means provided by docker.
+
+Then you need to change the `firmware-action` configuration to include the image reference (digest hash).
+
+```admonish example
+~~~json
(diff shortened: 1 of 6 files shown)


docs/src/firmware-action/troubleshooting.md
```diff
@@ -0,0 +1,42 @@
+# Troubleshooting common problems
+
+Many `firmware-action` errors and warnings come with suggestion on how to fix them.
+
+Other than that, here some common problems and solutions.
```
MDr164, line 5:
```suggestion
Other than that, here are some common problems and solutions.
```

docs/src/firmware-action/troubleshooting.md
```diff
@@ -0,0 +1,42 @@
...
+
+
+```admonish tip
+The first thing when troubleshooting is to look thought the output for errors and warnings. Many of these messages come with `suggestion` with instructions on possible solution.
```
MDr164, line 9:
```suggestion
The first thing when troubleshooting is to look through the output for errors and warnings. Many of these messages come with a `suggestion` with instructions on possible solutions.
```

docs/src/firmware-action/offline_usage.md
```diff
@@ -0,0 +1,33 @@
...
+~~~json
+"sdk_url": "ghcr.io/9elements/firmware-action/coreboot_4.19:main"
+~~~
+```
+
+However, in this configuration `firmware-action` (or rather `dagger`) will always connect to the internet and download the manifest to see if a new container needs to be downloaded. This applies to all tags (`main`, `latest`, `v0.8.0`, and so on).
+
+If you need to use `firmware-action` offline, you have to first acquire the container. Either by running `firmware-action` at least once online, or by other means provided by docker.
+
+Then you need to change the `firmware-action` configuration to include image reference (digest hash).
```
MDr164, line 15:
```suggestion
Then you need to change the `firmware-action` configuration to include the image reference (digest hash).
```

MDr164 (approved):

PR #454, opened 2024-12-02
feat(action): check for any undefined environment variable in config

### LINKED ISSUE #384: warn if JSON configuration contains undefined environment variables
Since JSON configuration does allow for use of environment variables, we should warn user when there is environment variable that is not defined or does not exist.

Right now when we have something like:
```json
"sdk_url": "ghcr.io/9elements/firmware-action/uroot_${UROOT_VERSION}:main"
```
It simply results into:
```json
"sdk_url": "ghcr.io/9elements/firmware-action/uroot_:main"
```

And this is not always easy to debug. There should be at the very least a warning.

AtomicFS:
- until now it was rather hard to debug when the issue was one or more missing environment variables
- with this patch, all environment detected variables in the JSON configuration file must be defined at the time of execution (can be empty)

Golang's [os.ExpandEnv](https://pkg.go.dev/os#ExpandEnv) does not perform such check:
> References to undefined variables are replaced by the empty string

fixes #384

### DIFF
--- a/action/recipes/config.go
+++ b/action/recipes/config.go
@@ -11,6 +11,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"dagger.io/dagger"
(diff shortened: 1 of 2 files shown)


action/recipes/config.go
```diff
@@ -20,7 +21,10 @@ import (
 )
 
 // ErrVerboseJSON is raised when JSONVerboseError can't find location of problem in JSON configuration file
```
MDr164, line 23: Can you move the comment for the error into the `var()` block over the error it is describing? Right now it looks like it is a description of both the `ErrVerboseJSON` and `ErrEnvVarUndefined` objects.

MDr164 (approved):

PR #460, opened 2024-12-04
feat: detect changes in sources

### LINKED ISSUE #441: rebuild module when sources changed
Right now I have to specify in Taskfile:
```yaml
...
    sources:
      - .. all of the sources ...
    generates:
      - output-coreboot/coreboot.rom
    method: timestamp
...
```

Especially the `sources` part is annoying. I have to basically list all files that are already in the firmware-action JSON config. This should be build-in.

We already have the `generates` functionality covered since #166 . Although we still need to add `--buldozer-mode`.

AtomicFS:
- few fixes (mostly small stuff)
- speed up filesystem.DirTree (I just happened to stumble upon it)
- add functions to detect changes in files
  - add new filesystem functions which can detect changes in files base on time-stamps
  - since we work with large projects, such as Linux Kernel, this detection should be as fast as possible
  - to make it fast, we solely rely on time, no hashing
  - the idea is that firmware-action will create a time-stamp file at the end of each execution (for each target)
  - at the start of each execution firmware-action will load time-stamp file from previous execution and check if any file in sources is newer that the saved time-stamp
  - the detection function is lazy, and will return on first positive match
  - if the no file in sources is newer, target should be considered up-to-date and skipped
- add the ability to detect changes in sources
- in addition, the dagger setup was moved so now the loop for up-to-date module is much faster
- prettify summary table

Summary before:
```
[INFO   ] Build summary:
uroot-example: Skipped
linux-example-with-uroot: Skipped
coreboot-example-with-linuxboot: Skipped
```

Summary now:
```
[INFO   ] Build summary:
+---------------------------------+------------+
| MODULE                          | STATUS     |
+---------------------------------+------------+
| uroot-example                   | Up-to-date |
| linux-example-with-uroot        | Up-to-date |
| coreboot-example-with-linuxboot | Up-to-date |
+---------------------------------+------------+
```

fixes #441

### DIFF
--- a/action/filesystem/filesystem.go
+++ b/action/filesystem/filesystem.go
@@ -232,7 +232,7 @@ func AnyFileNewerThan(path string, givenTime time.Time) (bool, error) {
 	if errors.Is(err, ErrPathIsDirectory) {
 		errMod := filepath.WalkDir(path, func(path string, info os.DirEntry, _ error) error {
 			// skip .git
-			if info.Name() == ".git" {
+			if info.Name() == ".git" && info.IsDir() {
 				return filepath.SkipDir
 			}
 			if !info.IsDir() {
--- a/action/filesystem/filesystem.go
+++ b/action/filesystem/filesystem.go
@@ -126,7 +126,11 @@ func MoveFile(pathSource, pathDestination string) error {
 func DirTree(root string) ([]string, error) {
 	var files []string
 
-	err := filepath.Walk(root, func(path string, info os.FileInfo, _ error) error {
+	// WalkDir is faster than Walk
+	// https://pkg.go.dev/path/filepath#Walk
+	//   > Walk is less efficient than WalkDir, introduced in Go 1.16, which avoids
+	//   > calling os.Lstat on every visited file or directory.
+	err := filepath.WalkDir(root, func(path string, info os.DirEntry, _ error) error {
 		foundItem := path
 		if info.IsDir() {
 			foundItem = fmt.Sprintf("%s/", path)
--- a/action/filesystem/filesystem.go
+++ b/action/filesystem/filesystem.go
@@ -7,8 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/plus3it/gorecurcopy"
 )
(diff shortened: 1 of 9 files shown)


AtomicFS:
@MDr164 I just made one small change - when there is no `timestamp` file (the `givenTime` is zero), it behaves like the target it is `up-to-date`. This is mostly for convenience in CI.

`up-to-date` just means that is the output directory exists, it will not re-build.

AtomicFS:
No to toot my own horn, but I like that this is actually much faster than `Taskfile`. They go the slow way file-by-file, which this does the directory at the time. And with minimal reads too thanks to the `WalkDir` (improvement over `Walk`).

AtomicFS:
@MDr164 do you have time for review?

action/filesystem/filesystem.go
```diff
@@ -141,3 +149,122 @@ func DirTree(root string) ([]string, error) {
...
+}
+
+// SaveCurrentRunTime writes the current time into file
+func SaveCurrentRunTime(pathLastRun string) error {
+	// Create temporaryFilesDir
+
+	// Create directory if needed
+	dir := filepath.Dir(pathLastRun)
+	err := CheckFileExists(dir)
+	if errors.Is(err, os.ErrNotExist) {
```
MDr164, line 191: You can probably save yourself the checks here as `os.MkdirAll` will simply return `nil` if the directory already exists and do nothing in that case.

action/recipes/recipes.go
```diff
@@ -132,31 +138,78 @@ func Build(
...
 // Execute a build step
+// func Execute(ctx context.Context, target string, config *Config, interactive bool, bulldozeMode bool) error {
 func Execute(ctx context.Context, target string, config *Config, interactive bool) error {
-	// Setup dagger client
-	client, err := dagger.Connect(ctx, dagger.WithLogOutput(os.Stdout))
+	// Prep
+	_, err := os.Stat(TimestampsDir)
 	if err != nil {
-		return err
+		err = os.MkdirAll(TimestampsDir, os.ModePerm)
```
MDr164, line 174: Same here, you can save yourself the call to `os.Stat` as `os.MkdirAll`  does not return an error on an existing directory

MDr164 (commented):
Looks mostly fine, just some minor comments

action/filesystem/filesystem.go
AtomicFS: Valid point.

MDr164 (approved):

PR #467, opened 2024-12-11
feat: add universal module

### LINKED ISSUE #244: add universal module
It should be quite easy to add a universal module, kinda like `edk2` or `u-root` is at the moment - it just runs user-defined command is a docker container.

This way people could use `firmware-action` even for yet-unsupported projects, or even some non-firmware projects and steps.

Instead of:
```json
"build_command": "cmd"
```

it would use:
```json
"build_commands": [
  "cmd1",
  "cmd2",
]
```

AtomicFS:
fixes #244 

I just copied `u-root` and renamed stuff, removed unnecessary things, and changed the build command into slice instead of single string.

### DIFF
--- a/action/recipes/universal.go
+++ b/action/recipes/universal.go
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+
+// Package recipes / universal
+package recipes
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"dagger.io/dagger"
+	"github.com/9elements/firmware-action/action/container"
+)
+
+// UniversalSpecific is used to store data specific to universal
+// ANCHOR: UniversalSpecific
+type UniversalSpecific struct {
+	// Specifies build commands to execute inside container
+	BuildCommands []string `json:"build_commands" validate:"required"`
+}
+
+// ANCHOR_END: UniversalSpecific
+
+// ANCHOR: UniversalOpts
+
+// UniversalOpts is used to store all data needed to run universal
+type UniversalOpts struct {
+	// List of IDs this instance depends on
+	// Example: [ "MyLittleCoreboot", "MyLittleEdk2"]
+	Depends []string `json:"depends"`
+
+	// Common options like paths etc.
+	CommonOpts
+
+	// Universal specific options
+	UniversalSpecific
+}
+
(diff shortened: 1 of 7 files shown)


action/recipes/universal.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"dagger.io/dagger"
+	"github.com/9elements/firmware-action/action/container"
+)
+
+// UniversalSpecific is used to store data specific to universal
```
MDr164, line 15: I know there comments have just been slightly adapted from the u-root module but they read really weird now. Maybe something like this?
```suggestion
// UniversalSpecific is used to store data specific to the universal command module
```

action/recipes/universal.go
```diff
@@ -0,0 +1,99 @@
...
+type UniversalSpecific struct {
+	// Specifies build commands to execute inside container
+	BuildCommands []string `json:"build_commands" validate:"required"`
+}
+
+// ANCHOR_END: UniversalSpecific
+
+// ANCHOR: UniversalOpts
+
+// UniversalOpts is used to store all data needed to run universal
```
MDr164, line 26: Same here. Maybe:
```suggestion
// UniversalOpts is used to store all data needed to run universal commands
```

action/recipes/universal.go
```diff
@@ -0,0 +1,99 @@
...
+func (opts UniversalOpts) GetDepends() []string {
+	return opts.Depends
+}
+
+// GetArtifacts returns list of wanted artifacts from container
+func (opts UniversalOpts) GetArtifacts() *[]container.Artifacts {
+	return opts.CommonOpts.GetArtifacts()
+}
+
+// buildFirmware builds universal
```
MDr164, line 51: Also here but not sure what to change it to

MDr164 (commented):
Code-wise looking good, left some comments about the docs/comments being a bit odd on this one

action/recipes/universal.go
AtomicFS: `universal command module` :D sounds like NASA project :D

But yeah, you are right, they sound odd, but I had no idea what to change it to. `universal module` just sounds a but weird.

MDr164 (approved):

PR #468, opened 2024-12-11
ci(lint): markdown link check as warning

AtomicFS:
- downgrade failed markdown link checks from error to warning
- this is because of frequent errors


MDr164 (approved):
Sounds like a reasonable change, that lint check shouldn't block merging

PR #476, opened 2024-12-14
test(action): add timeout to SSH server start

AtomicFS:
When I was looking into #470, I had a issue when the tests would take forever, especially the `TestOpenSSH`. It would hang for 1 hour until the `go test ... -timeout 60m ...` timeout would cause the test to fail.

So I made this timeout feature. I set it only to 5 seconds, but it should be fine I think.


action/container/ssh_test.go
```diff
@@ -101,6 +102,13 @@ func TestOpenSSH(t *testing.T) {
 			myContainer, err := Setup(ctx, client, &tc.opts, "")
 			assert.NoError(t, err)
 
+			// Set timeout for opening the SSH server
+			timeout := make(chan bool, 5)
+			go func() {
```
MDr164, line 107: The more idiomatic approach is to create a 5 second ticker and check if a tick occurred in the select statement. Afterwards stop the ticker. Saves you the manual goroutine with a timeout channel.

MDr164 (approved):
We can ignore the linter warning for now as it's just the indirect dep one

PR #483, opened 2024-12-19
build(deps): bump all golang dependencies

AtomicFS:
manually run:
```
go get -u
go mod tidy
```


AtomicFS:
Looks maybe like issue in dagger? No idea.

MDr164:
`go get -u` without a package (aka updating all of them) is prone to cause issues because it will update the indirect deps as well regardless of which version was pulled in by the direct deps. If the indirect ones get too new their API might change and therefore break code in the direct deps. I always refrain from a global `go get -u` and rather just call it only on the direct deps. So https://github.com/9elements/firmware-action/pull/482 should be enough to update our Go deps properly.

AtomicFS:
Fair enough. I was just too lazy to debug #482, so I wanted to try this approach just in case it would work.

AtomicFS:
Closed in favor of #482

PR #484, opened 2024-12-19
chore(deps): bump x/net due to an upstream CVE

MDr164:
This should silence the linters and dependabot


AtomicFS (approved):
Looks simple enough

PR #491, opened 2025-01-06
fix(ci): stop scheduled docker container builds

AtomicFS:
- I think that the scheduled runs might be messing up the containers in the GitHub registry
- we have a problem that multiple containers are not possible to download, error says "manifest unknown"
- this might be the cause, but not sure yet
- it should be fine to build the containers only once and never touch them even again


MDr164 (approved):
Either way should be fine to not build that many potentially equal images

PR #495, opened 2025-01-08
feat: coreboot git describe

AtomicFS:
The motivations behind this:
- coreboot likes to embed its source version into the compiled binary
- coreboot does that by calling `git describe ...`
- however with firmware-action we can assume people will include coreboot as git submodule, and thus when compiling coreboot in container the git commands will fail because of missing `.git` directory
- to fix this we will run manually the `git describe` command and then pass it in form of env variable into the container

The coreboot part `crebootPassEnvVars` is not tested now. I am too tired to look into unit-testing that right now (the supporting functions have unit tests).

I just ran the `firmware-action` on one project and tested it manually - it seems to work exactly as expected. Not sure if it is worth it to write unit-test.

@MDr164 if you think unit-test for the `crebootPassEnvVars` is a good idea, then I will make it tomorrow.

**TODO:**
- [x] Check for existence of `.coreboot-version` which should take precedence before auto-generated env var (should prevent creation of env var). Might be good idea to make a list of priorities in the comments.
- [ ] I will also look into implementing this (read as copy-pasting) for Linux kernel too.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -46,6 +46,7 @@
     "HEALTHCHECK",
     "Hadolint",
     "IPXE",
+    "KERNELVERSION",
     "NOTSET",
     "REPOPATH",
     "TARGETARCH",
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
 .coverage
-.task
 .example_runs
+.task
+bin/
 cmd/firmware-action/coverage.html
 cmd/firmware-action/coverage.out
-bin/
+cmd/firmware-action/recipes/__tmp_files__/
+docker/coreboot/coreboot-*
 megalinter-reports
 node_modules
-cmd/firmware-action/recipes/__tmp_files__/
 output-*/
-docker/coreboot/coreboot-*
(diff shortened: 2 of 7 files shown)


MDr164:
Now the wrong Marv is even tagged on GitHub and not only on slack anymore, great :grin:

AtomicFS:
Oh, sorry, sleep deprivation :D

I will stop for today, probably doing more damage then anything else :D

MDr164:
> btw, what do you think about the unit-testing?

Gotta take a closer look later, sadly I'm a bit busy this week

> The biggest hurdle I think would be the need for `cbfstool`, so that we can actually check the embedded version string.

That is part of coreboot and could be precompiled inside the container if we need it. Some distributions even package it already.

> The commands to check are:
> 
> ```
> cbfstool output-coreboot-RELEASE/coreboot.rom extract -n build_info -f /tmp/foo
> grep COREBOOT_VERSION /tmp/foo
> ```

Doesn't look too hard and could be automated, no?

MDr164 (approved):

AtomicFS:
I was working on the unit test :D

But I guess we can merge that in separate PR :D

PR #505, opened 2025-01-15
test: coreboot version passing

AtomicFS:
Tests for #495

### DIFF
--- a/cmd/firmware-action/recipes/coreboot_test.go
+++ b/cmd/firmware-action/recipes/coreboot_test.go
@@ -5,10 +5,12 @@ package recipes
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"testing"
 
 	"dagger.io/dagger"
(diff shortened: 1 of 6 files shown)


cmd/firmware-action/recipes/coreboot_test.go
```diff
@@ -57,7 +60,98 @@ func TestCorebootProcessBlobs(t *testing.T) {
 	}
 }
 
-func TestCoreboot(t *testing.T) {
+func gitCloneWithCache(dirName string, destination string, branch string, tag string, depth int, url string, fetch bool) error {
```
MDr164, line 63: I'm wondering if this function should use `t.Helper()` to omit any errors here from the line counting inside the actual test functions for debugging. Though not sure if that makes that much sense here. Also the amount of input parameters here is a bit high, what do you think about passing an options struct instead?

cmd/firmware-action/recipes/coreboot_test.go
```diff
@@ -170,6 +291,311 @@ func TestCoreboot(t *testing.T) {
...
+				version := string(versionEntryPatter.FindSubmatch(corebootVersionFileContent)[1])
+
+				versionPattern := regexp.MustCompile(tc.versionRegex)
+				assert.True(t, versionPattern.MatchString(version), fmt.Sprintf("found version '%s' does not match expected regex '%s'", version, tc.versionRegex))
+			}
+		})
+	}
+}
+
+func gitCloneAsSubmoduleWithCache(projectName string, dirName string, destination string, tag string, url string, fetch bool) error {
```
MDr164, line 317: Same as the other non-test function in this file

cmd/firmware-action/go.mod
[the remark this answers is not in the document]
AtomicFS: Well, technically, we can remove it. You found a old remnant of me trying to copy directories recursively.

Initially I wanted to use `os.CopyFS` which is available since `go v1.23`, but then found out that it cannot deal with symbolic links.

And then I found that we already have a function for this exact thing :D

So technically speaking, we can remove this requirement on `go v1.23`.

cmd/firmware-action/recipes/coreboot_test.go
AtomicFS: I like both suggestions, the `t.Helper()` and passing in struct.

AtomicFS:
I have amended away the `go v1.23` change. The rest of the changes are now in separate `refactor` commit for easier review.

cmd/firmware-action/recipes/coreboot_test.go
```diff
@@ -57,7 +60,109 @@ func TestCorebootProcessBlobs(t *testing.T) {
...
+	dirName     string
+	destination string
+	branch      string
+	tag         string
+	depth       int
+	url         string
+	fetch       bool
+}
+
+func gitCloneWithCache(t testing.TB, opts *gitCloneOpts) {
```
MDr164, line 74: I'm not fully sure but I think `testing.TB` should be passed as a pointer here and not by value

MDr164 (commented):
Looks pretty solid, just one minor item

cmd/firmware-action/recipes/coreboot_test.go
AtomicFS: Yes, I think you are correct.

AtomicFS: Hmmm ... in [examples](https://gopherguides.com/golang-fundamentals-book/07-testing/helpers) I do see it as passed by value

AtomicFS: I will change it to pointers and see if anything breaks.

AtomicFS: Oh, it already is a pointer :D

AtomicFS: It already is a pointer. For example test `TestCorebootBuild(t *testing.T)` calls `gitCloneWithCache(tb testing.TB, opts *gitCloneOpts)` with following arguments `gitCloneWithCache(t, &opts)`, where `t` is already a pointer.

I have renamed the `t` to `tb` to differentiate a little bit better.

MDr164: I haven't looked at examples I just checked the gopkg docs and `Helper()` is a method defined as `func (T*) Helper()` so works on pointers but given that TB is just an interface it will accept a pointer if said pointer implements it so I think it should be fine as is.

MDr164 (approved):

PR #508, opened 2025-01-20
fix(docker): downgrade NodeJS for udk2017

### LINKED ISSUE #500: edk2 container `udk2017` fails to build
After merging #498 (fixing issue #494) the edk2 container `udk2017` fails to build because of version mismatch:
```
25  : │ [12.6s] | The following packages have unmet dependencies:
25  : │ [12.6s] |  nodejs : Depends: libc6 (>= 2.28) but 2.27-3ubuntu1.6 is to be installed
25  : │ [12.7s] | E: Unable to correct problems, you have held broken packages.
```

Before the #498, container shipped with NodeJS v8.10.0 (support ended 2019-12-31).

AtomicFS:
- NodeJS v20 requires libc6 version >= 2.28, but that is not possible with this old Ubuntu
- switching to newer Ubuntu is not a option either because of how fragile and snowflaky edk2 is
- updating libc6 is just a terrible idea, that means that udk2017 cannot run NodeJS v20
- ~~the best we can do is to run v18, which is end-of-life soon, but that is the best we can do~~
- **UPDATE: the best we can do v16, everything else is too new for this old Ubuntu**
- still, it is much better than v10 that the Ubuntu Bionic (18.04) ships with

fixes #500

### DIFF
--- a/docker/compose.yaml
+++ b/docker/compose.yaml
@@ -56,6 +56,7 @@ services:
         - PYTHON_VERSION=python2
         - GCC_CROSS_COMPILER_PACKAGES_ARM=gcc-aarch64-linux-gnu gcc-arm-linux-gnueabi gcc-i686-linux-gnu
         - GCC_CROSS_COMPILER_PACKAGES_X86=gcc-i686-linux-gnu
+        - NODESOURCE=https://deb.nodesource.com/setup_16.x
       # NonOSI wasn't used in 2017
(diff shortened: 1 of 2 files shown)


MDr164 (dismissed):

AtomicFS:
OK, NodeJS `v16` it is. That is the newest we can use.

PR #519, opened 2025-01-23
feat(cmd): change BuildCommand into slice

### LINKED ISSUE #518: refactor edk2 `BuildCommand` into slice of strings
I think it would make it much nicer if the `BuildCommand` were a slice, the commands would then be much easier to read. Unfortunately this is a breaking change, but not that significant, people should be able to just change the JSON:

```patch
- "build_command": "source ./edksetup.sh; build -a X64 -p UefiPayloadPkg/UefiPayloadPkg.dsc -b DEBUG -t ${GCC_TOOLCHAIN_VERSION}",
+ "build_command": ["source ./edksetup.sh; build -a X64 -p UefiPayloadPkg/UefiPayloadPkg.dsc -b DEBUG -t ${GCC_TOOLCHAIN_VERSION}"],

```

That is if they do not want to take advantage of this new feature.

This new feature would also mean that we have to drop edk2's `DefconfigPath`.

@MDr164 when parsing JSON (or YAML #446), is the order of elements in JSON array preserved when fed into go's array/slice? My commons sense says it should be preserved, I just want to confirm.

AtomicFS:
- this is a breaking change
- this should make the JSON configuration files much more readable for edk2
- this change means we have to drop DefconfigPath for edk2

fixes #518


AtomicFS:
Failing because of
```
error: input: container.from resolve: failed to resolve image ghcr.io/9elements/firmware-action/coreboot_4.19:main: failed to resolve source metadata for ghcr.io/9elements/firmware-action/coreboot_4.19:main: failed to copy: httpReadSeeker: failed open: content at https://ghcr.io/v2/9elements/firmware-action/coreboot_4.19/manifests/sha256:b0beef78d1d9a2398e5909c6ccb92416eccb1aa7fc33e60ea442493b0de14d00 not found: not found
```

AtomicFS:
Now when I am thinking about this, it is inconvenient that stuff like `source` or `export`ed variable will not work across elements in the slice. Maybe on each step we could export the `env` and the import it before the next command?

I don't know ... not sure if this is a good idea anymore.

MDr164 (commented):
So the removed DefconfigPath for EDK2 is now part of the BuildCommand field or how does that work in practice?

AtomicFS:
> So the removed DefconfigPath for EDK2 is now part of the BuildCommand field or how does that work in practice?

I did not know in which order the PR's will get merged ;)

AtomicFS:
Oh sorry, I misread the comment. Yes, the general idea is to do not use the `DefconfigPath` for edk2. It was Patrick's idea if I am not mistaken. It is to store build flags and arguments for edk2 build. Very hacky non-standard solution.

I believe that it was meant to simplify local builds.

If we would switch to `BuildCommand` as slice, then the whole JSON config would be more readable and then I would not see any good reason for this `DefconfigPath`.

However the issue is that `exported` variables or `source`ed files will not work across elements in the slice. But having the commands separately would be much nicer for debugging with interactive `dagger` debug, I think.

AtomicFS:
Oh, I just had a thought ... not ideal, but would make this change possible.

We could have a `BuildCommand` slice and just write it to a file, with shebang at the beginning, and executing it. That way we would not have to worry about the loss of context (environment variables, ... etc). And the JSON file would be cleaner and more readable.

@MDr164 do you think it is a good idea, or should we just scrap this whole PR.

AtomicFS:
We agreed to scrap this, and instead instruct the use in documentation to write a shell scripts. This would be much less painful.

So in documentation we should describe how to do that.

PR #526, opened 2025-01-27
chore(docker): add new coreboot containers

AtomicFS:
I assume this fails now on the `auto-rerun-failed` job because the `docker-build-and-test-rerun.yml` workflow is not yet in `main`.

```
HTTP 404: Not Found (https://api.github.com/repos/9elements/firmware-action/actions/workflows/docker-build-and-test-rerun.yml)
Error: Process completed with exit code 1.
```

I quickly made [my own fork](https://github.com/AtomicFS/firmware-action) to test this on `main` branch and it seems to work just fine. So I think it will work once on `main`. See [action run 13201854560](https://github.com/AtomicFS/firmware-action/actions/runs/13201854560).

I pinned the reusable workflow / composite action `userdocs/gh-cli-workflow-reruns` to a specific commit instead of just `main` to mitigate potential supply-chain attack, since it is not that widely used and does not have that many eyes on it. I think that `dependabot`s `package-ecosystem: 'github-actions'` should keep it up-to-date, so we can actually get notified of any changes.


AtomicFS:
I do not know why, but the container builds fail from time to time, and re-running them seems to fix the problem.

MDr164 (approved):
looks fine, let's see if this fixes the weird fetching issue

AtomicFS:
This one is fixing the segmentation fault when building the containers. The fetching should be fixed with #552

PR #527, opened 2025-01-27
chore(docker): add new linux containers and remove old

AtomicFS:
- remove non-LTS linux containers
- note about discontinuation into README
- bump LTS releases to newer patch versions
- add new 6.12 LTS release


AtomicFS:
Linux examples fail because the new Linux containers are not yet published.

MDr164:
The CI still tries to run the removed containers so you'll have to adapt that part as well

AtomicFS:
Yeah, I am trying to split it into 2 PRs.

AtomicFS:
I renamed the linux defconfigs to new names, but to keep the old examples working, I had to also keep the old defconfigs. So to keep the git history for those files, I decided to `git mv` them, and then just make symbolic links. These links will then be removed in the next PR.

MDr164 (approved):

PR #528, opened 2025-01-27
chore(docker): add new edk2-stable202411 container

AtomicFS:


AtomicFS:
the `u-boot` example is failing because in the PR it detects no changes in the golang code, so it downloads `firmware-action` executable from releases, where the latest does not yet have `u-boot` support.

MDr164 (approved):

PR #532, opened 2025-01-28
ci(labeler): improve labeling for modules

AtomicFS:


.github/labeler.yml
```diff
@@ -47,29 +47,40 @@ module/coreboot:
...
+      - head-branch:
+          - 'stitching'
 module/u-root:
   - any:
       - changed-files:
           - any-glob-to-any-file:
               - 'cmd/firmware-action/recipes/uroot*'
               - 'docker/uroot/**'
+      - head-branch:
+          - 'uroot'
```
MDr164, line 82: Is that on purpose to have u-root here twice?

AtomicFS: Yes, it is on purpose, because sometimes I call it `u-root`, sometimes `uroot` (looks better IMHO in `dash-case` or `kebab-case`).

PR #536, opened 2025-01-30
ci(docker): stop removing old containers

AtomicFS:
- currently have a issue with Docker being unable to pull some containers
- maybe there is a bug in this step for removing old contianers causing problems, we could try to temporarily remove it

### DIFF
--- a/.github/workflows/container-cleanup.yml
+++ b/.github/workflows/container-cleanup.yml
@@ -1,0 +1,71 @@
+---
+# Remove old containers
+
+name: container cleanup
+on:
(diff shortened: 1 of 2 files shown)


MDr164 (commented):
Can you just comment that part of the yaml file out, given that we are not yet sure if that is actually the issue?

AtomicFS:
Fair enough. How about I move it to separate file?

MDr164:
> Fair enough. How about I move it to separate file?

Sounds good

PR #538, opened 2025-01-30
ci(docker): increase git commit sha slug to 12 characters

AtomicFS:
- we are still having some issues with Docker pulling containers, maybe increasing the short git commit sha form default 8 characters to 12 will improve things (it will certainly decrease the likelihood of collision)


AtomicFS:
Now that I am digging into the docker issue, there might be different reason for this.

AtomicFS:
Never-mind, the thing I was thinking of is seems like red herring.

MDr164 (approved):

PR #551, opened 2025-02-04
feat(cmd): expose building contianers on the fly

### LINKED ISSUE #163: add option to build docker container on the fly
It would be nice for development purposes (as debugging) to have the option to call firmware-action and use Dockerfile instead of URL. There is already a lot of work already in place to allow this, but it is not finished and exposed to the user.

AtomicFS:
- for a while now there was a functionality to build Docker containers on the fly from Dockerfile, but this feature was not exposed to the user
- this patch exposes the feature
- it will not work will with our Dockerfiles out of the box because they are meant to be used with accompanying Docker Compose file, so using one of our Dockerfiles would require to edit it (just replace default ARG values with desired ones)

fixes #163

### DIFF
--- a/cmd/firmware-action/container/container.go
+++ b/cmd/firmware-action/container/container.go
@@ -89,14 +89,28 @@ func (opts SetupOpts) Validate() error {
 }
 
 // Setup for setting up a Docker container via dagger
-func Setup(ctx context.Context, client *dagger.Client, opts *SetupOpts, dockerfileDirectoryPath string) (*dagger.Container, error) {
-	// dockerfileDirectoryPath allows to use Dockerfile and build locally,
-	//   which is handy for testing changes to said Dockerfile without the need to
-	//   have the container uploaded into package registry
-
+func Setup(ctx context.Context, client *dagger.Client, opts *SetupOpts) (*dagger.Container, error) {
 	err := opts.Validate()
 	if err != nil {
 		return nil, err
+	}
+
+	// dockerfileDirectoryPath allows to use Dockerfile and build locally,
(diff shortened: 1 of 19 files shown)


cmd/firmware-action/container/container.go
```diff
@@ -89,14 +89,28 @@ func (opts SetupOpts) Validate() error {
...
+	err := opts.Validate()
+	if err != nil {
+		return nil, err
+	}
+
 	// dockerfileDirectoryPath allows to use Dockerfile and build locally,
 	//   which is handy for testing changes to said Dockerfile without the need to
 	//   have the container uploaded into package registry
+	dockerfileDirectoryPath := ""
+	dockerfilePathPattern := regexp.MustCompile(`^file:\/\/.*`)
```
MDr164, line 102: Does this path pattern also work under non-Linux systems actually?

MDr164 (commented):
Looks good but as this changes the public API this change should technically be followed by a bump of the version I assume.

AtomicFS:
> Looks good but as this changes the public API this change should technically be followed by a bump of the version I assume.

Since it is `feature`, if would bump version from `v0.12.x` to `v0.13.0`.

cmd/firmware-action/container/container.go
AtomicFS: Well, the `file://` string is there only to detect is it not URL but filepath. It only serves for detection and is the removed. It never goes further than `if-else` statement under it.

AtomicFS: see `pathPattern` and subsequent `pathPattern.ReplaceAllString`

MDr164 (approved):

PR #556, opened 2025-02-06
feat: publish to dockerhub

### LINKED ISSUE #552: publish docker containers also to DockerHub
We should publish only containers on releases. We should not clutter the DockerHub with containers from `main` branch. Maybe this will help with #549.

Also, not uploading containers from `main` helps because I do not know how to automatically clean DockerHub of old images. Not uploading them in the first place makes my life easier.

Containers should be published to https://hub.docker.com/repositories/9elementscyberops

AtomicFS:
Now we can publish to both GitHub and DockerHub

fixes #552

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -79,6 +79,7 @@
     "distclean",
     "dockerfile",
     "dockerfiles",
+    "dockerhub",
     "dooper",
     "dotconfig",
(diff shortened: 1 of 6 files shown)


MDr164 (approved):
lgtm but I'm also not the best at reading python code

PR #564, opened 2025-02-11
ci: better caching for coreboot

AtomicFS:
- we are getting 443 from review.coreboot.org when multiple builds run in short succession
- these changes should improve caching and reduce load we put on coreboot infrastructure

I changed caches in examples so that we do not get conflicts. Technically speaking the content could be re-used but `actions/cache` are finecky. Not only that they have to match the `key`, but also the `paths`, and in `example.yml` we have named the coreboot repositories `./my_super_dooper_awesome_coreboot`, while in `docker-build-and-test.yml` we call it just `coreboot`.


AtomicFS:
@MDr164 ping

MDr164 (approved):

PR #569, opened 2025-02-12
feat(cmd): refactor how coreboot blobs are handled

AtomicFS:
- this is rather drastic change to handling blobs in coreboot
- coreboot can have far more blobs that we currently support and the current setup does not scale
- we remove all of the hard-coded stuff and replace it with much more flexible setup where use has to define key-value map for the blobs
- this is a breaking change, which will require users to update the JSON configuration file
- I categorized it as feature to bump minor version in next release, rather than refactor which would bump only patch


MDr164 (commented):
lgtm, still need to run a test locally with it

AtomicFS:
@MDr164 I tried it on project and found some issue with detecting if blob is directory or file (possibly because wrong working directory or something). I did not bother to investigate, I just moved the check closer to the location where it is actually needed. And it fixed the issue.

MDr164 (approved):

PR #574, opened 2025-02-13
fix(docker): compilation of coreboot utils

AtomicFS:


AtomicFS:
@MDr164 i think it is ready for review. Sorry for the spam, I miss-clicked and accidentally not made it into draft.

PR #588, opened 2025-02-19
feat: support multiple configs

### LINKED ISSUE #559: allow use of multiple configuration files
If would be nice if people could split their configuration into multiple files, as use it then like:

```
firmware-action build --config="fa-01.json" --config="fa-02.json" ...
```

The structures would have to be merged before validation, because people might have cross-config-file dependencies.

AtomicFS:
fixes #559

### DIFF
--- a/cmd/firmware-action/main.go
+++ b/cmd/firmware-action/main.go
@@ -12,6 +12,7 @@ import (
 	"log/slog"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/9elements/firmware-action/cmd/firmware-action/filesystem"
 	"github.com/9elements/firmware-action/cmd/firmware-action/logging"
@@ -42,7 +43,7 @@ var CLI struct {
 	Indent bool `default:"false" help:"enable indentation for JSON output"`
 	Debug  bool `default:"false" help:"increase verbosity"`
 
-	Config string `type:"path" required:"" default:"${config_file}" help:"Path to configuration file"`
+	Config []string `type:"path" required:"" default:"${config_file}" help:"Path to configuration file, supports multiple flags to use multiple configuration files"`
 
 	Build struct {
 		Target    string `required:"" help:"Select which target to build, use ID from configuration file"`
@@ -76,7 +77,7 @@ func run(ctx context.Context) error {
 	)
 	slog.Info(
 		fmt.Sprintf("Running in %s mode", mode),
-		slog.String("input/config", CLI.Config),
+		slog.Any("input/config", CLI.Config),
 		slog.String("input/target", CLI.Build.Target),
 		slog.Bool("input/recursive", CLI.Build.Recursive),
 	)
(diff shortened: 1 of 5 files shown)


cmd/firmware-action/recipes/config.go
```diff
@@ -230,6 +230,78 @@ func (c Config) AllModules() map[string]FirmwareModule {
...
+		slog.Warn("Module already exists in configuration, and will be overwritten",
+			slog.String("module_name", key),
+			slog.String("suggestion", "Make sure each module has unique name, not only per single configuration file but across all configuration files."),
+		)
+	}
+}
+
+// Merge method will take other Config instance and adopt all of its modules
+func (c Config) Merge(other Config) (Config, error) {
+	// Not sure of there is cleaner way to do this :/
```
MDr164, line 244: Maybe via a range loop and some type reflection but that might be a bit hard to read

MDr164 (approved):
I still think simply hashing would have been enough and that this is a bit over-engineered but it works so who am I to judge

AtomicFS:
I think you mean #592, that is the one for change detection.

cmd/firmware-action/recipes/config.go
AtomicFS: I am trying to re-write this to use reflection, since that would be far more flexible.

I stared with `AllModules` method since it should be practically the same thing, but a bit simpler.

```golang
// AllModules method returns slice with all modules
func (c Config) AllModules() map[string]FirmwareModule {
	modules := make(map[string]FirmwareModule)

	config := reflect.ValueOf(c)
	for i := range config.Type().NumField() {
		for module := range config.Field(i).Interface().(map[string]FirmwareModule) {
			slog.Error("REFLECTION",
				slog.Any("module", module),
			)
		}
	}

	...

}
```

This code does not work, ends in panic.

```
panic: interface conversion: interface {} is map[string]recipes.CorebootOpts, not map[string]recipes.FirmwareModule [recovered]
	panic: interface conversion: interface {} is map[string]recipes.CorebootOpts, not map[string]recipes.FirmwareModule
```

Which I guess makes sense, but only `for module := range config.Field(i).Interface()` will not work since type is missing. So how do I do this @MDr164 ?

AtomicFS: Solved, ChatGPT to the rescue :D

cmd/firmware-action/main.go
```diff
@@ -189,7 +190,7 @@ func parseCli() (string, error) {
 
 	case "generate-config":
 		// Check if config file exists
-		err := filesystem.CheckFileExists(CLI.Config)
```
MDr164, line 192: Might be worth making sure `CLI.Config` is at least of size 1. I don't think there are any situations right now where that is not the case but can't hurt to check

cmd/firmware-action/recipes/config.go
```diff
@@ -206,30 +207,84 @@ type Config struct {
...
-		modules[key] = value
-	}
-	for key, value := range c.Universal {
-		modules[key] = value
-	}
-	for key, value := range c.UBoot {
-		modules[key] = value
+
+	configValue := reflect.ValueOf(c)
+	// configType := reflect.TypeOf(c)
```
MDr164, line 212: Please remove the commented out code :pleading_face: :point_right: :point_left:

MDr164 (commented):
Reading the reflection code gave me an aneurysm but it looks like it is working as intended :+1:

cmd/firmware-action/main.go
AtomicFS: Added check for length

cmd/firmware-action/recipes/config.go
AtomicFS: Removed

MDr164 (approved):

PR #589, opened 2025-02-19
fix(cmd): golang test for linux module

AtomicFS:
- we changed naming of Linux containers, but forgot to update the golang tests


AtomicFS:
@MDr164 ping

MDr164 (approved):

PR #592, opened 2025-02-20
feat: support change detection in config

### LINKED ISSUE #543: detect changes in JSON config file to trigger re-builds
Changes in the JSON configuration should trigger re-build of respective module. It should not trigger re-build of all modules.

Since JSON can contain environment variables, we have to take that into account. The steps should be:
- load the JSON config file
- fill in environment variables
- save the evaluated JSON config into somewhere into `.firmware-action` next to `timestamps`, maybe into separate directory called `configs`
- next time use JSON in `.firmware-action/configs` to detect if module needs to be re-built

AtomicFS:
This PR adds the ability to firmware-action to detect changes in the JSON configuration file and re-build only the affected module (and possibly modules that depend on it).

And yeah, I know that simple checksum of the config file would do the job, but I just wanted to avoid running build on modules that do not need it. It will make `firmware-action` just a bit faster this way.

I did consider to save into a file only the small struct related to each module, but it was proving to be far too much work. So instead we save the whole configuration file (for each module separately) and only compare the relevant modules between these two configurations (between the snapshot saved in `./firmware-action` and current config).

~~There is no test for this. At least not yet. Should I make a the test for this? I am kinda lazy. But I guess I should ... :sleepy:~~

OK, I made a test for this. It is not very comprehensive, but I think it is good enough.

fixes #543 
~~needs first #559~~


cmd/firmware-action/filesystem/filesystem.go
```diff
@@ -265,3 +270,31 @@ func AnyFileNewerThan(path string, givenTime time.Time) (bool, error) {
...
+	//   https://stackoverflow.com/a/31976060
+
+	// ASCII characters (printable and non-printable)
+	//   https://en.wikipedia.org/wiki/Control_character#In_ASCII
+	charactersASCII := regexp.MustCompile(`[<>:"/\\|?*\x00-\x1F\x7F]`)
+	result := charactersASCII.ReplaceAllString(name, "_")
+
+	// unicode control characters
+	//   https://www.compart.com/en/unicode/category/Cc
+	unicodeControlChars := regexp.MustCompile("[\u0000-\u001f\u007F-\u009f]")
```
MDr164, line 286: This could probably just be `strings.ToValidUTF8()`

cmd/firmware-action/filesystem/filesystem.go
```diff
@@ -265,3 +270,31 @@ func AnyFileNewerThan(path string, givenTime time.Time) (bool, error) {
 	// If path is neither file nor directory
 	return false, err
 }
+
+// Filenamify converts a string to a valid and safe filename
+func Filenamify(name string) string {
```
MDr164, line 275: Might be worthwhile to take the name and extension as separate arguments to make sure the extension isn't being cut off

cmd/firmware-action/recipes/config.go
```diff
@@ -294,9 +299,21 @@ type FirmwareModule interface {
...
+// ===============================
+//  Functions for FirmwareModules
+// ===============================
+
+// FilenameForFirmwareModule is used to take a user-defined module name and make it into filename, removing
+// all problematic characters
+func FilenameForFirmwareModule(name string, extension string) string {
+	// For example:
+	//   "Coreboot Example" should return "Coreboot_Example.json"
+	return fmt.Sprintf("%s.%s", filesystem.Filenamify(name), extension)
```
MDr164, line 311: We already have the new `Filenamify` function, would be nice to sanity check the extension string as well then so that we don't accidentally input something like the whole bee movie script here.

cmd/firmware-action/recipes/recipes.go
```diff
@@ -199,19 +209,55 @@ func Execute(ctx context.Context, target string, config *Config) error {
...
+			// If the config is old / obsolete and no longer valid, it should just be ignored
+			// and it should be assumed that re-build is needed
+			if err != nil {
+				changedConfig = true
+				slog.Warn(
+					fmt.Sprintf("The configuration used for previous build, stored in '%s', is not valid and will be assumed obsolete", CompiledConfigsDir),
+				)
+			} else {
+				oldModules := oldConfig.AllModules()
+				changedConfig = !cmp.Equal(modules[target], oldModules[target])
```
MDr164, line 231: We could probably use `reflect.DeepEqual` here but `cmp.Equal` should be fine as well

AtomicFS: I will switch to `reflect.DeepEqual` in separate commit, so we can easily revert if things go south.

MDr164: We can also just keep the extra dependency here. I don't really mind for this project which one we use.

cmd/firmware-action/filesystem/filesystem.go
```diff
@@ -10,9 +10,12 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/plus3it/gorecurcopy"
+	"tailscale.com/util/truncate"
```
MDr164, line 18: When I said "why not use truncate" yesterday I incorrectly has the "Trim" functions from the strings pkg in mind but none of them operate on length so the approach you had previously is actually better. I'm not sure if it is worth it pulling in a full VPN library that happens to have a truncate function just to shorten a string. Almost feels like the JS package "isTrue" :D

AtomicFS: Oh, yeah, I was looking at that but when I was reading the description I thought I must have mis-remembered and after some searching found this library :D

AtomicFS: OK, I reverted back to the original.

MDr164 (approved):

PR #603, opened 2025-02-27
Feat/GitHub ci better composite2

AtomicFS:


AtomicFS:
If you will have time @MDr164 , could you try these changes? @PatrickRudolph feel free to also test this out ;)

I have did some experiments in [firmware-action-example](https://github.com/9elements/firmware-action-example/) directory. Namely just running `go-task build:uroot` bunch of times while:
- changing the `coreboot-linuxboot-example.json` config file (for example changing `container_input_dir` for `uroot` changes nothig, but will trigger rebuild) to test the Config change detection
- `touch ./coreboot-linuxboot-example/u-root/README.md` to test the TimeStamp detection
- checking out different commit in the `u-root` repo to test the Git commit hash detection (specifically checking out commit `62f8e804`, which is right after the `v0.14.0` tag and changes nothing substantial).

From my experiments it seems to work just as intended.

MDr164 (commented):
Nice! I did not spot any errors and it seems to be doing what we expect it to. Given that this is a fairly large change there might be some oversights on my end. Would be nice to have another pair of eyes on this, otherwise I'll approve this change latest by end of the week if no one else feels like testing this :)

AtomicFS:
/remind @MDr164 to approve this on March 14

MDr164 (approved):
Works as expected here

Issue #549, opened 2025-02-03
docker fails to pull containers


AtomicFS:
`docker pull` fails to fetch older containers (older = sometimes even less than 3 days old). I can reproduce this on my machine, and I see this also in GitHub's CI.

I have no idea why is this happening.

The only known workaround for this right now is to manually trigger a `dagger` workflow for given tag.

@MDr164 mentioned that his setup with `podman` instead of `docker` does not suffer the same issues. Hence why we opened issue #541 

# Steps to reproduce
```
docker pull ghcr.io/9elements/firmware-action/coreboot_4.19:v0.10.1
```
```
docker manifest inspect ghcr.io/9elements/firmware-action/coreboot_4.19:v0.10.1
```

# Attempts to fix this
- #538 
- #536 
- #556

AtomicFS:
Seems like the issue was resolved one way or another.

PR #618, opened 2025-03-14
fix: container cleanup

AtomicFS:
I have not seen or heard any problems related #549, so I think we turn the cleanup on again, but this time as separate workflow (see #536).

- run weekly
- also keep discontinued containers clean

Discontinued containers did get rather messy, for `linux_6.1.45` we had over 30 un-tagged images (not anymore since I ran this manually).

I do see warnings on `coreboot_4.20` and `coreboot_4.22`, but that is because there is no such package / container anymore (I suspect that we switched to `coreboot_4.20.1` and `coreboot_4.22.01` faster than we make releases, and maybe overzealous cleanup removed them? Not sure. But it is only warning on discontinued container, so I don't really care.

### DIFF
--- a/.github/workflows/container-cleanup.yml
+++ b/.github/workflows/container-cleanup.yml
@@ -2,7 +2,11 @@
 # Remove old containers
 
 name: container cleanup
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'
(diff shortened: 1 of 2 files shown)


MDr164 (approved):
Let's hope it works as expected and doesn't bring any weird side-effects that aren't documented for GHCR and such :eyes:

PR #634, opened 2025-03-19
fix(action): upload artifact only for specified target

### LINKED ISSUE #628: automatic uploading of artifacts should not upload all of them
Right now the artifact upload will upload all artifacts, even those that have been built and downloaded before.

This is because the `copy_artifacts` job in `action.yml` file will move all instances of `output_dir` (if exists) to be artifacted.

AtomicFS:
fixes #628


MDr164 (approved):
Sad snowball noises

PR #654, opened 2025-04-01
feat: goreleaser tweaks

### LINKED ISSUE #653: bug: load `iptable_nat` kernel module
### Description

`firmware-action` needs kernel module `iptable_nat` to be loaded in order for [Dagger to work](https://docs.dagger.io/troubleshooting/#dagger-restarts-with-a-cni-setup-error).

We already have a detection of this problem and we print error message #455 , but we could make this simpler on the users. For ArchLinux we already have [AUR package](https://aur.archlinux.org/packages/firmware-action) where we could configure this, and we could use `goreleaser` to also generate packages for Ubuntu / Debian / Fedora.

### Version of firmware-action

v0.16.0

### Environment

local build

### CPU architecture

x86_64, arm64, i386

### OS

Linux

### Reproduction URL

_No response_

### Reproduction steps

```bash
Try to run without `iptable_nat` loaded.
```

### Screenshots

```bash

```

### Logs

```bash

```

### JSON configuration

```json

```

AtomicFS:
I hope this works. I made a `Taskfile` to locally test things.
- making some goreleaser defaults explicit
- add nFPM to goreleaser
  - add config to automatically load linux kernel modules
- add Taskfile

fixes #653

### DIFF
--- a/cmd/firmware-action/.goreleaser.yaml
+++ b/cmd/firmware-action/.goreleaser.yaml
@@ -46,3 +46,33 @@ archives:
     format_overrides:
       - goos: windows
         formats: ['zip']
+
+nfpms:
+  # Docs: https://goreleaser.com/customization/nfpm/
+  - id: packages
+    package_name: 'firmware-action'
+    file_name_template: >-
+      {{- trimsuffix .ConventionalFileName .ConventionalExtension -}}
+      {{- if and (eq .Arm "6") (eq .ConventionalExtension ".deb") }}6{{ end -}}
+      {{- if not (eq .Amd64 "v1")}}{{ .Amd64 }}{{ end -}}
+      {{- .ConventionalExtension -}}
+    vendor: '9elements GmbH'
+    homepage: 'https://github.com/9elements/firmware-action'
+    maintainer: 'AtomicFS'
+    description: |
+      Build system for open source firmware with unified builds across
+      development and CI environments
+    license: 'MIT'
+    formats:
+      - deb
+      - rpm
+      - archlinux
+    replaces:
+      - golang-github-fti
+      - golang-fti
+    bindir: /usr/bin
+    section: default
+    priority: extra
+    contents:
+      - src: ./kernel-modules-load.conf
+        dst: /lib/modules-load.d/firmware-action.conf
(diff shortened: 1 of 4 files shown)


cmd/firmware-action/.goreleaser.yaml
```diff
@@ -46,3 +46,33 @@ archives:
...
+    description: |
+      Build system for open source firmware with unified builds across
+      development and CI environments
+    license: 'MIT'
+    formats:
+      - deb
+      - rpm
+      - archlinux
+    replaces:
+      - golang-github-fti
```
PatrickRudolph, line 71: need to be adapted

cmd/firmware-action/.goreleaser.yaml
```diff
@@ -12,14 +12,27 @@ dist: '../bin'
...
-  - env:
+  - id: build
+    env:
       - CGO_ENABLED=0
     goos:
       - linux
       - windows
       - darwin
+    goarch:
+      - 386
```
MDr164, line 23: Do we really want to build for this arch? I doubt we could easily support that one in  case issue ever arise from it.

cmd/firmware-action/kernel-modules-load.conf
```diff
@@ -0,0 +1 @@
+iptable_nat
```
MDr164, line 1: What do we need this one for?

AtomicFS: Dagger cannot run without it. #455

cmd/firmware-action/.goreleaser.yaml
AtomicFS: fair enough

cmd/firmware-action/.goreleaser.yaml
AtomicFS: Oh yeah, forgot that. Thanks

cmd/firmware-action/kernel-modules-load.conf
AtomicFS: At the end of the `.goreleaser.yaml` there is this `contents` entry, where we take this file `kernel-modules-load.conf` and put it into `/lib/modules-load.d/firmware-action.conf`

cmd/firmware-action/.goreleaser.yaml
AtomicFS: FYI, `386` was there by default ;)

MDr164 (approved):
lgtm

PR #662, opened 2025-04-10
refactor(cmd): use T.Chdir for cleaner tests

### LINKED ISSUE #612: Migrate to better testing context and Chdir functions
Golang in `v1.24` added new `T.Context` and `B.Context` methods (see [golang 1.24 release notes](https://tip.golang.org/doc/go1.24#:~:text=the%20new%20t.context%20and%20b.context%20methods%20return)), which are cool and would make everything nicer.

Golang `v1.24` was released recently (February 2025) so we should wait until it gets into main-stream distributions like `Ubuntu`, `Fedora`, and so on. But some distributions might take too long.

So, since we can already get `v1.24` in GitHub as essentially default, then lets not wait any longer.

## Ubuntu
According to [packages.ubuntu.com](https://packages.ubuntu.com/search?suite=default&section=all&arch=amd64&keywords=golang-go&searchon=names), at the moment:
- Ubuntu `25.04` (`Plucky`) has `v1.24`
- Ubuntu `24.10` (`Oracular`) has `v1.23`
- Ubuntu `24.04 LTS` (`Noble`) has `v1.22`

## GitHub defaults
While on GitHub `ubuntu-latest` means `ubuntu-24.04` (at the time of writing), action `actions/setup-go@v5` will already set-up `v1.24` since it is considered `stable` (if used with `go-version: 'stable'` input, which I think would be default good practice).

## Fedora
Fedora 41 (at the time the latest release) was released with Golang `v1.23` and so far seems to be sticking with it ([pkgs.org](https://pkgs.org/download/golang)).

Next release of Fedora, the 42, is scheduled for 2025-04-22.

## Debian
Debian 12 (`Bookworm`) is right now at Golang `v1.19`.

Debian Sid sits on `v1.24`.

AtomicFS:
- T.Chdir was introduced in new golang v1.24
- it makes test much cleaner

fixes #612

### DIFF
--- a/cmd/firmware-action/container/container_test.go
+++ b/cmd/firmware-action/container/container_test.go
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MIT
 
+//go:build go1.24
+
 // Package container
 package container
 
@@ -147,12 +149,8 @@ func TestSetup(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			// Create test files
-			pwd, err := os.Getwd()
-			defer os.Chdir(pwd) // nolint:errcheck
-			assert.NoError(t, err)
 			tmpDir := t.TempDir()
-			err = os.Chdir(tmpDir)
-			assert.NoError(t, err)
+			t.Chdir(tmpDir)
 
 			// Create InputDirs
 			for _, val := range tc.opts.InputDirs {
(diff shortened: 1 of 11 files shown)


MDr164 (approved):
Nice! I'm still surprised the linter requires the build constraints to force go 1.24 as we already set go 1.24 in the go.mod file.

PR #693, opened 2025-05-09
ci: docs and tests

AtomicFS:
Adding a job to merge the coverage reports. Not sure what to do about it afterwards.


MDr164:
> Not sure what to do about it afterwards.

We could upload the report to codecov for example

AtomicFS:
> We could upload the report to codecov for example

I was thinking the same, but I am not sure how to use codecov (yet).

PR #694, opened 2025-05-13
ci: more linting

AtomicFS:
This PR essentially just enables more linters and does some cosmetic changes to the code as reported by the newly enabled linters.

TLDR:
- make golangci more strict
- enable all linters by default and explicitly disable some
- this work was prompted with discussion about ambiguous unicode characters and invisible characters
  - initially wanted to only enable `asciicheck` and `bidichk`, but went do the rabbit-hole of enabling more linters


cmd/firmware-action/recipes/coreboot_test.go
```diff
@@ -641,7 +604,7 @@ func TestCorebootSubmodule(t *testing.T) {
 			version := string(versionEntryPatter.FindSubmatch(corebootVersionFileContent)[1])
 
 			versionPattern := regexp.MustCompile(tc.versionRegex)
-			assert.True(t, versionPattern.MatchString(version), fmt.Sprintf("found version '%s' does not match expected regex '%s'", version, tc.versionRegex))
```
MDr164, line 644: This line here should not work though CI still runs. :crab:

AtomicFS: You are correct, it should use [assert.Truef](https://pkg.go.dev/github.com/stretchr/testify/assert#Truef)

MDr164 (approved):
I'd say the new config is reasonable

PR #698, opened 2025-05-19
fix(cmd): error on nested output

AtomicFS:
- users can try to use nested output directories or duplicate paths, which is not supported and will cause problems
- the configuration validation step now checks for this and reports error on violation
- when module is to be rebuilt, it's output directory is deleted, in case the outputs are nested, they get deleted too

```
Here are few examples:
  VALID:
    .
    ├── output-linux
    │   └── linux.bin
    └── output-uroot
        └── uroot.bin
  VALID:
    .
    └── output
        ├── output-linux
        │   └── linux.bin
        └── output-uroot
            └── uroot.bin
  INVALID:
    .
    └── output
        ├── output-linux
        │   └── linux.bin
        └── uroot.bin
```


mynetz:
@AtomicFS: As always, thank you very much for the quick investigation! Also the good abstract understanding of the issue :) 
I did a quick glance on the description: That seems to be pretty much the source of my problem and a good way to avoid it for the future.
The code seems also to match this. However, I think I can not (yet :)) do a full review.

AtomicFS:
@MDr164 ping

MDr164 (approved):

PR #732, opened 2025-08-19
feat: add tarfile support

### LINKED ISSUE #730: feat: Allow import/usage of local container image
I would like to use custom local container images in an offline build. Either via the Dockerfile feature or another way to prevent FWA (or Dagger) to query an online resource.

While the Dockerfile can be used directly in a FWA build, and works fine. There is no way to then to do something that is equivalent to "offline usage" as described for public containers in: https://9elements.github.io/firmware-action/firmware-action/offline_usage.html

It would be great to have the option to provide either a reference in the local Docker cache or a container image. Eg. `"sdk_url": "container_image.tar"` or `"sdk_url": "<my_local_label>:<tag>@sha256:...."`

**Note**: It seems there is a development in dagger that might bring something similar as a generic option to dagger.
- dagger/dagger#10737
- dagger/dagger#10810
It may take some time though.

AtomicFS:
fixes #730

- refactor mode detection - URL, Dockerfile, and Tarfile
- add support for Taskfile
- add tests for container `Setup` function to test all modes


AtomicFS:
No problem, I am happy that it works :)

We will then make a new release.

